PR #11, opened 2019-04-18
Pypy release and Docker container

fbertsch:


fbertsch:
@whd pinged for review for the CI deploy and container bits. We'll need to get the appropriate creds added for this repo.

.circleci/config.yml
```diff
@@ -33,8 +60,21 @@ workflows:
...
+      - deploy:
+          requires:
+            - test
+          filters:
+            tags:
+              only: /.*/
+            branches:
+              only: master
+
+  nightly: # Test nightly to test external dependencies (MPS, probe-info-service)
```
whd: This appears to be failing currently (but running, so that's good).

mozilla_schema_generator/configs/glean.yaml
```diff
@@ -5,179 +5,153 @@ metrics:
         not:
           - glean_ping_info
       type: boolean
-      labeled: false
```
whd: I have no idea what these bits are about but it looks like they may be the reason CI is failing.

fbertsch: Right, it is. PR passes so we should be good to go.

PR #13, opened 2019-04-24
Write pings to directories with proper name

fbertsch:


fbertsch:
This should now match what we see in mozilla-pipeline-schemas: `namespace`/`doctype`/`doctype`.`version`.schema.json

mozilla_schema_generator/__main__.py
```diff
@@ -56,13 +56,15 @@
 )
 def generate_main_ping(config, out_dir, split, pretty):
     schema_generator = MainPing()
+    if out_dir:
```
acmiyaguchi: It might be useful to mention somewhere that the `out_dir` corresponds to the schema namespace.

PR #18, opened 2019-04-25
Add script to generate and upload schemas

fbertsch:


fbertsch:
There's still a few TODOs:

- Will this script create a tarball and push that to GCS, or will jenkins see the new push to MPS and create its own?
- We need to get the proper credentials and bucket for GCS public repos
- We need to setup credentials to let this script push to MPS

fbertsch:
@whd, you seem most likely to have opinions on those todos listed above.

bin/schema_generator.sh
```diff
@@ -0,0 +1,75 @@
...
+
+# TODO: Update schema mapping for validation
+# TODO: Handle overwriting glean schemas
+
+GCS_BUCKET="mozilla-generated-schemas"
+MPS_BRANCH="generated-schemas"
+
+# 0. Install dependencies
+
+cargo install --git https://github.com/acmiyaguchi/jsonschema-transpiler --branch 0.2 --force
```
fbertsch: @acmiyaguchi is this still how the transpiler should be retrieved?

bin/metadata_merge
```diff
@@ -0,0 +1,57 @@
...
+    'schema',
+    type=click.Path(
+        dir_okay=False,
+        file_okay=True,
+        writable=True,
+        exists=True,
+    ),
+    required=True
+)
+def main(metadata, schema):
```
fbertsch: @jklukas is this how you envisioned metadata being merged?

bin/schema_generator.sh
acmiyaguchi: Yes, though this will change very soon. I'm moving the repository to the mozilla org either tomorrow or next week, and making master the default branch. I'm also trying to generate binaries for cross platform linux use, but the infrastructure for that isn't set up (and installing cargo isn't too bad anyways).

fbertsch: Great, keep us updated. The cross-platform binaries could be useful for MPS testing, but aren't required here.

bin/schema_generator.sh
```diff
@@ -0,0 +1,75 @@
...
+SCHEMAS_DIR="schemas"
+BASE_DIR="/Users/frankbertsch/repos/sandbox"
+
+cp metadata_merge.py $BASE_DIR
+cd $BASE_DIR
+rm -rf mozilla-pipeline-schemas
+
+git clone https://www.github.com/mozilla-services/mozilla-pipeline-schemas
+cd mozilla-pipeline-schemas/$SCHEMAS_DIR
+git checkout ping-metadata
```
fbertsch: When we merge the GCS metadata, we'll checkout `dev` instead.

whd:
>     * Will this script create a tarball and push that to GCS, or will jenkins see the new push to MPS and create its own?

The latter.

>     * We need to get the proper credentials and bucket for GCS public repos

Absent a compelling reason, this should also happen on jenkins (as triggered by the MPS push).

>     * We need to setup credentials to let this script push to MPS

I will work on setting this up for airflow.

bin/schema_generator.sh
```diff
@@ -0,0 +1,75 @@
...
+
+# 7. Push to branch of MPS
+# Note: This method will keep a changelog of releases.
+# If we delete and newly checkout branches everytime,
+# that will contain a changelog of changes.
+
+cd ../
+
+rm -rf templates tests validation
+find . -not -name "*.schema.json" -type f | xargs git add
```
whd: Does this flow correctly do nothing when no changes have been made to the schemas?

bin/schema_generator.sh
```diff
@@ -0,0 +1,75 @@
...
+cd $BASE_DIR
+rm -rf mozilla-pipeline-schemas
+
+git clone https://www.github.com/mozilla-services/mozilla-pipeline-schemas
+cd mozilla-pipeline-schemas/$SCHEMAS_DIR
+git checkout ping-metadata
+
+# 2. Remove all non-json schemas (e.g. parquet)
+
+find . -not -name "*.schema.json" -type f | xargs rm
```
whd: Is this necessary? I can image a future in which we have a unified branch (master) that is updated by both humans and machines, where parquet schemas are simply ignored in the GCP pipeline.

bin/metadata_merge
jklukas: Yes, this looks entirely reasonable.

bin/schema_generator.sh
fbertsch: Not really necessary, but I would prefer to only have the relevant schemas available. Why don't we add the parquet schemas back in later if they become useful?

fbertsch: Basically my goal is to have an easy way of saying "what is running on prod"? And the parquet schemas are not included in that (for GCP).

Dockerfile
```diff
@@ -19,18 +19,31 @@ RUN apt-get update && \
...
 
 WORKDIR ${HOME}
+RUN mkdir ${HOME}/bin
+ADD ./bin ${HOME}/bin
+ENV PATH $PATH:${HOME}/bin
 
 USER ${USER_ID}
+
+RUN git config --global user.name "Generated Schema Creator"
+RUN git config --global user.email "telemetry-alerts@mozilla.com"
```
whd: Now that we've generated creds for this user, this should be `dataops+pipeline-schemas@mozilla.com`. Still TBD how we're going to pass in the credentials.

bin/schema_generator.sh
```diff
@@ -0,0 +1,75 @@
...
+# that will contain a changelog of changes.
+
+cd ../
+
+rm -rf templates tests validation
+find . -not -name "*.schema.json" -type f | xargs git add
+git checkout $MPS_BRANCH || git checkout -b $MPS_BRANCH
+git commit -a -m "Auto-push from schema generation"
+
+git push --repo https://name:password@bitbucket.org/name/repo.git
```
whd: Per https://bugzilla.mozilla.org/show_bug.cgi?id=1547333#c2 I think the easiest thing to do here is to set the ssh key via an env var and have this script write the contents of that var to `~/.ssh/id_rsa`.

bin/schema_generator.sh
whd: This is fine for now, since the generated branch will be gcp-only. I'll want to revisit this in the hypothetical future where we have a single branch. Since the responsibility of this script will not include uploading to GCS the deletion of parquetmr files could still happen farther down the pipeline.

bin/schema_generator.sh
acmiyaguchi: I moved the repo to the mozilla org and [published a package](https://crates.io/crates/jsonschema-transpiler). This can now be `cargo install jsonschema-transpiler`.

Dockerfile
haroldwoo: ref https://bugzilla.mozilla.org/show_bug.cgi?id=1547333

bin/schema_generator.sh
fbertsch: I've updated this, but testing locally I still needed to input a password. The only way I could force push to github was using a personal access token.

Dockerfile
fbertsch: Updated.

bin/schema_generator.sh
fbertsch: Great news! Updating now.

fbertsch:
I've added a whitelist, and fixed the cherry-picking. Testing led to this branch:

https://github.com/mozilla-services/mozilla-pipeline-schemas/tree/generated-schemas

However, I've yet to get it to work pushing using the ssh auth.

bin/schema_generator.sh
```diff
@@ -0,0 +1,105 @@
...
+done
+
+# 6. Push to branch of MPS
+# Note: This method will keep a changelog of releases.
+# If we delete and newly checkout branches everytime,
+# that will contain a changelog of changes.
+
+cd ../
+
+rm -rf templates tests validation
```
fbertsch: We could consider removing the remaining files in the git repository, e.g. `CMakeLists.txt`, `Dockerfile`, etc. They aren't necessarily useful in the read-only git branch.

bin/schema_generator.sh
haroldwoo: im not sure about bitbucket, but github requires pushing to the ssh url rather than https url for password less

fbertsch: I did make that change, see the updated commit.

whd:
I've tested the generated-schemas branch on the jenkins infra and it mostly appears to work. The following issues still arise:

The `.bq` schemas aren't valid. This is a combination of https://github.com/mozilla/jsonschema-transpiler/issues/51 and (I think) the top-level `fields` being unnamed. @acmiyaguchi told me he's fixing at least the former of these.

The whitelist appears to work, but as a result a very limited subset of bq schemas are generated (e.g. no activity stream, webpagetest etc). Was this for testing purposes, or is bq table generation for less fancy schemas supposed to be incorporated into dev directly? Perhaps a blacklist would be better.

bin/schema_generator.sh
whd: I couldn't find an emoji for "this seems reasonable for a generated read-only git branch".

fbertsch:
@whd I added the whitelist so we could do a slow roll-out of schemas, while keeping the branch to contain only the ones that are deployed. All BQ table generation should be done during this build step.

bin/schema_generator.sh
fbertsch: I've confirmed that there is no commit pushed when there are no changes.

whd:
@fbertsch ok, can you whitelist activity-stream and eng-workflow (bmobugs only)? These (and core v9/10) are the current tables we're generating in GCP so having them in the first round of automated schemas would be ideal.

bin/schema_generator.sh
```diff
@@ -0,0 +1,102 @@
...
+# 2. Remove all non-json schemas (e.g. parquet)
+
+find . -not -name "*.schema.json" -type f | xargs rm
+
+# 3. Generate new schemas
+
+mozilla-schema-generator generate-glean-ping --out-dir . --pretty
+mozilla-schema-generator generate-main-ping --out-dir ./telemetry --pretty --split
+
+# 3a. Keep only whitelisted schemas
```
jklukas: "blacklist" and "whitelist" are going the way of "slave"/"master" terminology. If we can avoid them, I think we should. I've found "allowlist" and "blocklist" to be suitable alternatives.

jklukas:
> can you whitelist activity-stream and eng-workflow

I want to be careful here, especially with activity-stream, that we don't accidentally recreate tables and drop existing data. As long as the logic here doesn't drop tables, this should be okay.

I've already updated the activity-stream, bmobugs, and core tables "by hand" to include the new metadata fields (and to relax the previously required metadata.document_id to nullable, since it is no longer being sent). These may throw errors when we try to update schema since they contain old metadata fields not in the json schemas.

After I validate that the new metadata fields are flowing properly, I plan to update these tables to drop the old fields (tomorrow (Wednesday), if all goes well).

whd:
> > can you whitelist activity-stream and eng-workflow
> 
> I want to be careful here, especially with activity-stream, that we don't accidentally recreate tables and drop existing data. As long as the logic here doesn't drop tables, this should be okay.

The hand-crafted tables have a `_dev` suffix on them, whereas the managed ones do not, no tables are dropped.  A separate config change needs to be applied at https://github.com/mozilla-services/cloudops-infra/blob/ingestion/projects/beam/tf/modules/resources/dataflow_jobs/bigquery.tf#L30 to point the dataflow jobs at the new managed tables.

jklukas:
Good point that we're transitioning to the non-dev namespaces now. I think this sounds good. Once we change the job to point at the new namespaces, I can work on backfilling the `dev` activity-stream tables to the new ones.

fbertsch:
@whd I've updated the whitelist and made the BQ output conform to expectations. You can check the current output at https://github.com/mozilla-services/mozilla-pipeline-schemas/tree/generated-schemas

bin/schema_generator.sh
fbertsch: Good catch - I had never realized this. Updated!

whd:
Ok, I'm now seeing:

    * google_bigquery_table.telemetry_core_v5: googleapi: Error 400: The field specified for time partitioning can only be of type TIMESTAMP or DATE. The type found is: STRING., invalid
    * google_bigquery_table.telemetry_core_v1: 1 error(s) occurred:

which looks accurate since we're partitioning on `submission_timestamp`.

fbertsch:
@whd unfortunate, it doesn't look like we support the `date-time` format in JSON schemas yet in the transpiler. I'll open an issue.

whd:
@fbertsch can we also keep all the json schemas in the generated branch i.e. have the allowlist only apply to bq schema generation? This more accurately reflects how the current pipeline is running.

fbertsch:
@whd can do, should we keep pioneer schemas?

whd:
@fbertsch no need, we can decided later how to handle those.

fbertsch:
@whd addressed the feedback. The `submission_timestamp` fields are now `DATETIME`, so we may be able to test deploy.

bin/schema_generator.sh
```diff
@@ -0,0 +1,118 @@
...
+find ./telemetry -type f -exec metadata_merge $telemetry_metadata {} ";"
+find . -path ./telemetry -prune -o -type f -exec metadata_merge $structured_metadata {} ";"
+
+rm -rf $metadata_dir
+
+# 5. Add transpiled BQ schemas
+
+find . -type f -name "*.schema.json"|while read fname; do
+    BQ_OUT=${fname/schema.json/bq}
+    jsonschema-transpiler --type bigquery $fname | jq '.fields' > $BQ_OUT
```
acmiyaguchi: This no longer has to be run as of https://github.com/mozilla/jsonschema-transpiler/pull/61, it may actually cause the resulting schemas to be empty.

bin/schema_generator.sh
```diff
@@ -0,0 +1,156 @@
...
+    pip install -e ./mozilla-schema-generator
+}
+
+function clone_and_configure_mps() {
+    # Checkout mozilla-pipeline-schemas and changes directory to prepare for
+    # schema generation.
+
+    rm -rf mozilla-pipeline-schemas
+
+    git clone git@github.com:mozilla-services/mozilla-pipeline-schemas.git
```
whd: Might be good to make the git URL a top-level variable like the rest of the `MPS_` envs.

bin/schema_generator.sh
```diff
@@ -0,0 +1,156 @@
...
+    # shellcheck disable=SC1091
+    source msg-venv/bin/activate
+    pip install -e ./mozilla-schema-generator
+}
+
+function clone_and_configure_mps() {
+    # Checkout mozilla-pipeline-schemas and changes directory to prepare for
+    # schema generation.
+
+    rm -rf mozilla-pipeline-schemas
```
whd: Generally I would prefer something like `[[ -d mozilla-pipeline-schemas ]] && rm -rf mozilla-pipeline-schemas` but this is fine.

bin/schema_generator.sh
```diff
@@ -0,0 +1,156 @@
...
+
+    git clone git@github.com:mozilla-services/mozilla-pipeline-schemas.git
+    cd mozilla-pipeline-schemas/$MPS_SCHEMAS_DIR || exit
```
whd: What is this `|| exit` about?

bin/schema_generator.sh
```diff
@@ -0,0 +1,156 @@
...
+    git checkout $MPS_BRANCH_SOURCE
+
+    git branch -D $MPS_BRANCH_WORKING
+    git checkout -b $MPS_BRANCH_WORKING
+}
+
+function prepare_metadata() {
+    # Assumes that all schemas under the current directory are valid JSON schema.
```
whd: Might be worth simply adding `-name '*.schema.json`` to the find command(s).

bin/allowlist
```diff
@@ -0,0 +1,5 @@
+telemetry/core
```
whd: This file isn't an executable so in my opinion should not live in `bin`.

bin/schema_generator.sh
```diff
@@ -0,0 +1,156 @@
...
+
+    # Pioneer-study is not nested, remove it
+    rm -rf pioneer-study
+
+    # Replace newlines with backticks (hard to do with sed): cat | tr
+    # Remove the last backtick; it's the file-ending newline: rev | cut | rev
+    # Replace backticks with "\|" (can't do that with tr): sed
+    # Find directories that don't match any of the regex expressions: find
+    # Remove them: rm
+    < /app/mozilla-schema-generator/bin/allowlist tr '\n' '`' | \
```
whd: Maybe `$BASE_DIR` again. See previous comment on `allowlist` being a non-binary file. I _think_ this logic can be simplified to:

    find -name '*.bq' | grep -v -f $ALLOWLIST | xargs rm -f

bin/schema_generator.sh
```diff
@@ -0,0 +1,156 @@
...
+    # Remove metadata
+    rm -rf metadata
+
+    # Pioneer-study is not nested, remove it
+    rm -rf pioneer-study
+
+    # Replace newlines with backticks (hard to do with sed): cat | tr
+    # Remove the last backtick; it's the file-ending newline: rev | cut | rev
+    # Replace backticks with "\|" (can't do that with tr): sed
+    # Find directories that don't match any of the regex expressions: find
```
whd: The goal of this step is to remove bq schemas not in `allowlist`, we are keeping the directories and json schemas.

bin/schema_generator.sh
```diff
@@ -0,0 +1,156 @@
...
+    # Keep only the schemas dir
+    find .  -mindepth 1 -maxdepth 1 -not -name .git -exec rm -rf {} +
+    git checkout $MPS_BRANCH_WORKING -- schemas
+    git commit -a -m "Auto-push from schema generation"
+}
+
+function main() {
+    cd $BASE_DIR || exit
+
+    # -1. Setup ssh key and git config
```
whd: Why is this -1?

bin/schema_generator.sh
```diff
@@ -0,0 +1,156 @@
...
+    find . -not -name "*.schema.json" -type f -exec rm {} +
+
+    # 3. Generate new schemas
+    mozilla-schema-generator generate-glean-ping --out-dir . --pretty
+
+    # 4. Add metadata to all json schemas, drop metadata schemas
+    prepare_metadata
+
+    # 5. Add transpiled BQ schemas
+    find . -type f -name "*.schema.json"|while read -r fname; do
```
whd: nit: space for pipe

bin/schema_generator.sh
```diff
@@ -0,0 +1,156 @@
...
+    git checkout $MPS_BRANCH_PUBLISH || git checkout -b $MPS_BRANCH_PUBLISH
+
+    # Keep only the schemas dir
+    find .  -mindepth 1 -maxdepth 1 -not -name .git -exec rm -rf {} +
+    git checkout $MPS_BRANCH_WORKING -- schemas
+    git commit -a -m "Auto-push from schema generation"
+}
+
+function main() {
+    cd $BASE_DIR || exit
```
whd: What are these `|| exits` for? There's another one on line 151. We might want to just `set -e` if we're worried about commands failing for various reasons. In fact, we probably do want to `set -e`, but I haven't looked at where that would require changes to the script.

bin/schema_generator.sh
```diff
@@ -0,0 +1,156 @@
...
+        jsonschema-transpiler --type bigquery "$fname" > "$bq_out"
+    done
+
+    # 5b. Keep only allowed schemas
+    filter_schemas
+
+    # 6. Push to branch of MPS
+    cd ../ || exit
+    commit_schemas
+    git push --force
```
whd: From a technical perspective, force is not required for our use case (as we saw when accidentally committing to dev). I would recommend removing the force. In the rare case where somebody pushes the generated-schemas branch by hand in the middle of the airflow job's execution we would probably want the automated schema push to fail anyway. Maybe we have branch protection or something enabled now that requires this.

bin/schema_generator.sh
```diff
@@ -0,0 +1,156 @@
...
+    chown -R "$(id -u):$(id -g)" "$HOME/.ssh"
+    chmod 700 "$HOME/.ssh"
+    chmod 700 "$HOME/.ssh/id_ed25519"
+}
+
+function setup_dependencies() {
+    # Installs mozilla-schema-generator in a virtual environment
+
+    virtualenv msg-venv
+    # shellcheck disable=SC1091
```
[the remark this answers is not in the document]
acmiyaguchi: Shellcheck is a static analysis tool for bash scripts -- this line is to ignore not being able to source the activation script that's being used. 

https://github.com/koalaman/shellcheck/wiki/SC1091

bin/schema_generator.sh
acmiyaguchi: https://github.com/koalaman/shellcheck/wiki/SC2164

This is to prevent `cd` from the doing the wrong thing during a failure case.

bin/schema_generator.sh
acmiyaguchi: There are a couple of places where the folders and branches are being deleted without checking for existence. I can go back and add `set -eou pipefail` for errors, unset variables, and pipeline errors.

fbertsch:
Thanks for picking this up, Anthony! My only input is that the
jsonschema-transpiler dep wont be updated if the install is in the
dockerfile. I would recommend at least checking for an update in the script.

bin/allowlist
```diff
@@ -0,0 +1,5 @@
...
+org-mozilla-fenix/.*
+org-mozilla-reference-browser/.*
+activity-stream/.*
+eng-workflow/bmobugs
```
jklukas:
```suggestion
eng-workflow/bmobugs
mobile/activation
```

This new ping was just merged in https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/331 and I'd like to start testing it in the gcp pipeline tomorrow.

Dockerfile
```diff
@@ -15,22 +15,37 @@ RUN mkdir ${HOME} && \
...
+        file gcc libwww-perl jq && \
     apt-get autoremove -y && \
     apt-get clean
 
+# Install Google Cloud SDK
+RUN curl -sSL https://sdk.cloud.google.com | bash
+ENV PATH $PATH:$HOME/google-cloud-sdk/bin
+
 # Install Rust and Cargo
 RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain=${RUST_SPEC}
```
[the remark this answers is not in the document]
acmiyaguchi: Yeah, this is a perfectly legitimate way of installing rust. I've updated it to reflect the default arguments on the website: https://rustup.rs/

I don't think there are going to be any breaking changes to the language in the near future, so using stable should be fine.

bin/allowlist
acmiyaguchi: I agree, I'll move this to the project root. However, it seems like this should belong in the root of mozilla-pipeline-schemas.

bin/schema_generator.sh
acmiyaguchi: An original comment, I removed the numbering since the functions better define the scope of work.

bin/schema_generator.sh
acmiyaguchi: This looks like it does the right thing.

acmiyaguchi:
> Thanks for picking this up, Anthony! My only input is that the jsonschema-transpiler dep wont be updated if the install is in the dockerfile. I would recommend at least checking for an update in the script.

Unfortunately, there isn't a mechanism for updates. Instead, we would have to just force install over the original. We should just update the tag in the dockerfile when necessary. 

---

@whd r?

PR #21, opened 2019-05-20
1542709: Merge glean probes based on dependencies

mdboom:
This requires https://github.com/mozilla/probe-scraper/pull/87 to land first.

### DIFF
--- a/mozilla_schema_generator/glean_ping.py
+++ b/mozilla_schema_generator/glean_ping.py
@@ -13,20 +13,55 @@ from typing import List
 class GleanPing(GenericPing):
 
     schema_url = "https://raw.githubusercontent.com/mozilla-services/mozilla-pipeline-schemas/dev/schemas/glean/baseline/baseline.1.schema.json" # noqa E501
-    probes_url = "https://probeinfo.telemetry.mozilla.org/glean/{}/metrics"
+    probes_url_template = "https://probeinfo.telemetry.mozilla.org/glean/{}/metrics"
     repos_url = "https://probeinfo.telemetry.mozilla.org/glean/repositories"
 
-    default_probes_url = probes_url.format("glean")
+    default_dependencies = ['glean']
     default_pings = {"baseline", "events", "metrics"}
     ignore_pings = {"default", "glean_ping_info", "glean_client_info"}
 
     def __init__(self, repo):  # TODO: Make env-url optional
-        super().__init__(self.schema_url, self.schema_url, self.probes_url.format(repo))
+        self.repo = repo
+        super().__init__(
+            self.schema_url,
+            self.schema_url,
+            self.probes_url_template.format(repo)
+        )
+
+    def get_dependencies(self):
+        # Get all of the library dependencies for the application that
+        # are also known about in the repositories file.
+
+        # The dependencies are specified using library names, but we need to
+        # map those back to the name of the repository in the repository file.
+
+        dependencies = self._get_json(self.dependencies_url.format(self.repo))
+        dependency_library_names = list(dependencies.keys())
+
+        repos = self.get_repos()
+        repos_by_dependency_name = {}
+        for repo in repos:
+            for library_name in repo.get('library_names', []):
+                repos_by_dependency_name[library_name] = repo
+
+        dependencies = []
+        for name in dependency_library_names:
+            if name in repos_by_dependency_name:
+                dependencies.append(repos_by_dependency_name[name])
+
+        if not len(dependencies):
+            return self.default_dependencies
+
+        return dependencies
 
     def get_probes(self) -> List[GleanProbe]:
         probes = self._get_json(self.probes_url)
-        default_probes = self._get_json(self.default_probes_url)
-        items = list(probes.items()) + list(default_probes.items())
+        items = list(probes.items())
+        for dependency in self.get_dependencies():
+            dependency_probes = self._get_json(
+                self.probes_url_template.format(dependency)
+            )
+            items += list(dependency_probes.items())
 
         return [GleanProbe(_id, defn) for _id, defn in items]
 
--- a/mozilla_schema_generator/glean_ping.py
+++ b/mozilla_schema_generator/glean_ping.py
@@ -50,8 +50,10 @@ class GleanPing(GenericPing):
                 dependencies.append(repos_by_dependency_name[name])
 
         if not len(dependencies):
+            print(f"For {self.repo}, using default Glean dependencies")
             return self.default_dependencies
 
+        print(f"For {self.repo}, found Glean dependencies: {dependencies}")
         return dependencies
 
     def get_probes(self) -> List[GleanProbe]:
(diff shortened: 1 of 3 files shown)


mdboom:
Cc: @fbertsch, @acmiyaguchi, @Dexterp37

mozilla_schema_generator/glean_ping.py
```diff
@@ -13,20 +13,55 @@
...
+
+        repos = self.get_repos()
+        repos_by_dependency_name = {}
+        for repo in repos:
+            for library_name in repo.get('library_names', []):
+                repos_by_dependency_name[library_name] = repo
+
+        dependencies = []
+        for name in dependency_library_names:
+            if name in repos_by_dependency_name:
```
Dexterp37: Should we log something if this is not true?

mozilla_schema_generator/glean_ping.py
```diff
@@ -13,20 +13,55 @@
...
+                dependencies.append(repos_by_dependency_name[name])
+
+        if not len(dependencies):
+            return self.default_dependencies
```
Dexterp37: Should we log something here as well?

mozilla_schema_generator/glean_ping.py
mdboom: I don't think so.  Fenix has something like 45 dependencies and only one of them uses Glean.  If anything, I'd log *if* true.

mozilla_schema_generator/glean_ping.py
mdboom: I'd log maybe only if a `dependencies_url` was defined on the repo, (i.e. it's an app, not a library), and no dependencies were found.  That's something that should be resolved.  Libraries will never have to define this, even if they have dependencies, since the app returns a flattened tree of *all* of its dependencies.

mdboom: Looking at this some more, I don't think this is really an error case.  If reading the git hash of an application *before* the dependencies artifact is being emitted (as is currently the case with Fenix), that's not an error -- we just fall back to "Glean only" as a dependency.

Dexterp37: I think that, in general, I'm fine with having the fewest possible log lines. What I'd really like out of this is to speed up diagnosis of potential bugs: instead of my comments above, even something as easy as "only glean dependencies for project X" might work. It tells us straight away what to expect from the data :)

mdboom: That makes sense.  I've updated this to just log the dependencies found (and whether it's just using the defaults).

mdboom:
> Let's wait for the owners of the repo to review this, though, before merging!

Most definitely -- I'm just walking up here and don't really have a good sense of this repo as a whole :)

mozilla_schema_generator/glean_ping.py
```diff
@@ -13,20 +13,57 @@
...
+        )
+
+    def get_dependencies(self):
+        # Get all of the library dependencies for the application that
+        # are also known about in the repositories file.
+
+        # The dependencies are specified using library names, but we need to
+        # map those back to the name of the repository in the repository file.
+
+        dependencies = self._get_json(self.dependencies_url.format(self.repo))
```
fbertsch: Where is `dependencies_url` set?

fbertsch: Also, we may want to call it `dependencies_url_template` to match the `probes_url_template`

mozilla_schema_generator/glean_ping.py
```diff
@@ -13,20 +13,57 @@
...
+        dependency_library_names = list(dependencies.keys())
+
+        repos = self.get_repos()
+        repos_by_dependency_name = {}
+        for repo in repos:
```
fbertsch: Repos is a list of `(app_name, app_id)`. We can also retrieve `library_names` but we'll need to update the `get_repos` function below

mozilla_schema_generator/glean_ping.py
```diff
@@ -13,20 +13,57 @@
...
+            for library_name in repo.get('library_names', []):
+                repos_by_dependency_name[library_name] = repo
```
fbertsch: Do we want to use the `app_id` here? Do all libraries have those?

mozilla_schema_generator/glean_ping.py
```diff
@@ -13,20 +13,57 @@
...
+
+        dependencies = []
+        for name in dependency_library_names:
+            if name in repos_by_dependency_name:
+                dependencies.append(repos_by_dependency_name[name])
+
+        if not len(dependencies):
```
fbertsch: Will this ever happen? Shouldn't at least glean be a dependency?

fbertsch:
@mdboom let's see if we get a build triggered if you push to this branch. If not I'll have to figure out why circle isn't running

mozilla_schema_generator/glean_ping.py
mdboom: Ah -- I thought this was running, but it wasn't, since I was running the wrong command (small glitch in the docs, now fixed here).

mozilla_schema_generator/glean_ping.py
mdboom: Good point -- now that I actually have this running, this has been fixed to something that works...

mozilla_schema_generator/glean_ping.py
mdboom: I think we want to use name because that's what seems to go in the `probe_url_template` -- the idea is to make a list of URLs to probes from this...

mozilla_schema_generator/glean_ping.py
mdboom: If the application doesn't have any dependencies stored (such as when the dependencies artifact isn't being generated, as with the reference browser).  This is a test basically to make this fall back to having only glean as a dependency in that case.

mdboom:
Yeah -- not sure what's going on with circle here...  For me it doesn't even have a link to the Circle task (but that might be a permissions issue).

mozilla_schema_generator/glean_ping.py
fbertsch: If that's the case, won't it return early above (line 46?)

mdboom: That's a good point.  It probably can't (assuming the `repositories.yaml` file has everything it needs and the dependencies artifact is correct).  I'll take it out.

fbertsch:
@mdboom looks like the dependencies files are available now in the probe-scraper! I ran it on your branch here and ended with the following error:

```
  File "/Users/frankbertsch/repos/mozilla-schema-generator/mozilla_schema_generator/glean_ping.py", line 54, in get_dependencies
    repos_by_dependency_name[library_name] = repo.name
AttributeError: 'dict' object has no attribute 'name'
```

mdboom:
This is working locally for me now, and pulling the dependencies over HTTP as it should.

mdboom:
> LGTM! This was more straightforward than we initially thought it would be.

Yeah.  I dread the day if we ever have to track exactly which metrics came from which revision, but this tracking a superset of everything approach was fairly tractable.

PR #24, opened 2019-05-30
Add firefox installer pings to the allowlist

whd:
More generally I'm going to start filing PRs to add more entries to the allowlist, but this is a specific case that is important.

### DIFF
--- a/allowlist
+++ b/allowlist
@@ -4,3 +4,4 @@ org-mozilla-reference-browser/.*
 activity-stream/.*
 eng-workflow/bmobugs
 mobile/activation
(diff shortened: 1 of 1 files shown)


whd:
This didn't have the effect I was expecting (a BQ schema [here](https://github.com/mozilla-services/mozilla-pipeline-schemas/tree/generated-schemas/schemas/firefox-installer/install)). I'll investigate later today.

whd:
The problem will likely be resolved by https://github.com/mozilla/telemetry-airflow/pull/523/files.

PR #25, opened 2019-06-03
Remove addl properties

### LINKED ISSUE #22: Generated schemas should drop `additionalProperties` from filled-in sections
If a filled-in section currently does not have anything to fill in, it retains the `additionalProperties`. This leads to the transpiler including it as a MAP type, which means when it is eventually filled in with a field, the schema will change, which is not an allowed operation in BQ.

When we make this change, we'll also need to drop the existing Glean tables, because they have some fields as maps that shouldn't be currently.

fbertsch:
Fixes #22


mozilla_schema_generator/schema.py
```diff
@@ -34,11 +34,14 @@ class Schema(object):
     def __init__(self, schema: dict):
         self.schema = schema
 
-    def set_schema_elem(self, key: Tuple[str], elem: Any) -> dict:
+    def set_schema_elem(self, key: Tuple[str], elem: Any, *, propogate=True) -> dict:
```
acmiyaguchi: A  comment on `propagate` would be useful. It looks like it's the equivalent of `mkdir -p` for a nested folder where a parent doesn't exist.

tests/test_schema.py
```diff
@@ -47,3 +48,47 @@ def test_schema_size(self):
...
+                }
+            },
+            "type": "object"
+        }
+
+        res_schema.set_schema_elem(
+                ("properties", "b", "properties", "hello"),
+                {"type": "string"},
+                propogate=True)
+        print_and_test(expected, res_schema.schema)
```
acmiyaguchi: When `properties.b.properties.hello` is removed, does this turn back into `schema.schema`?

mozilla_schema_generator/generic_ping.py
```diff
@@ -94,12 +94,15 @@ def make_schemas(env: Schema, probes: List[Probe], config: Config,
...
-                False)
-            final_schema.delete_group_from_schema(schema_key + ("propertyNames",))
 
-        return schemas + [final_schema]
+        # Remove all additionalProperties (#22)
+        schemas.append(final_schema)
+        for s in schemas:
+            for key in config.get_match_keys():
+                s.delete_group_from_schema(key + ("propertyNames",), propogate=False)
+                s.set_schema_elem(key + ("additionalProperties",), False, propogate=False)
```
acmiyaguchi: Does this do the intended thing? It seems like you're removing the properties and leaving behind an additionalProperties field.

If I'm understanding correctly, this results in something like this:

```json
{
    "metrics": {
        "properties": {
            "my-probe": {
                "additionalProperties": false
            }
        }
    }
}
```

assuming a single probe in the config. This new bit of code is to avoid the following when probes are missing?

```json
{
    "metrics": {
        "additionalProperties": false
    }
}
```

fbertsch:
@acmiyaguchi generally the problem was that previously, if were filling this in:
```
{
    "metrics": {
        "type": "object",
        "additionalProperties": { "type": "integer" }
    }
}
```

then we would end up with (for e.g. 1 probe):
```
{
    "metrics": {
        "type": "object",
        "properties": {
            "probe-1": { "type": "integer" },
        }
        "additionalProperties": { "type": "integer" }
    }
}
```

And the transpiler would do the correct thing, and ignore the `additionalProperties` field, giving precedence to the struct type.

However, if there were no fill-in probes, we would end up with something like:
```
{
    "metrics": {
        "type": "object",
        "additionalProperties": { "type": "integer" }
    }
}
```

Which the transpiler would see as the pseudo-map type, resulting in a schema element of `STRUCT<key_value ARRAY<STRUCT<key STRING, value INT64>>>` (or similar). However, when we eventually did have probes to fill-in, the type changes and is not schema evolution compatible.

With this change, the result of no probes to fill-in would be:
```
{
    "metrics": {
        "type": "object",
        "additionalProperties": False
    }
}
```

...which, upon inspection, still returns this from the transpiler:
```
[
  {
    "mode": "NULLABLE",
    "name": "metrics",
    "type": "STRING"
  }
]
```

Hmm, it seems to me the transpiler should output an empty row here, do you agree? We'd need to check that that is acceptable for BQ.

tests/test_schema.py
fbertsch: Good question, it should if `propogate=True` for the deletion. I'll add this.

acmiyaguchi:
The type is a string because an empty object is not a well defined structure. There needs to be at least one field, even if it's just a dummy/placeholder value. I think instead, you should consider using the `propogate=True` method to drop the object entirely from the schema. We should also consider adding this as an option to the schema transpiler, since it might be desirable to omit rather than cast in this situation.

mozilla_schema_generator/schema.py
```diff
@@ -34,11 +34,21 @@ class Schema(object):
...
```
acmiyaguchi: `propogate` is a typo, it should be `propagate`.

fbertsch:
@acmiyaguchi that seems reasonable as well. I think an empty struct type would be most preferable (since it would match the expected schema of the ping) but deleting it would be second-best.

acmiyaguchi:
Looking at the BigQuery docs again, it looks like it is possible to have a [empty struct](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types). 

I also think that we might want to set `additionalProperties` to `true` instead of `false`, otherwise validation will definitely fail.

fbertsch:
I won't be using these for validation, currently we'll be using the generic glean schema for validation and the strict one for the BQ tables. We can update this later if that proves too lenient.

fbertsch:
Given that BQ does not support empty structs, we'll have to remove these from the schema instead.

fbertsch:
@acmiyaguchi this now deletes `additionalProperties` and propagates that deletion. RFAL

tests/test_schema.py
```diff
@@ -90,13 +90,6 @@ def test_set_elem_propogate(self):
...
+                propagate=True)
 
         print_and_test(expected, res_schema.schema)
-
-        # Deleting the elem again should match our original schema
-        res_schema.delete_group_from_schema(
-                ("properties", "b", "properties", "hello"),
-                propogate=True)
-
-        print_and_test(schema.schema, res_schema.schema)
```
acmiyaguchi: Did you accidentally remove this in the most recent commit?

fbertsch: Looks like it got lost in the rebase. I've added it back in.

PR #27, opened 2019-06-05
Glean ping schemas

fbertsch:
@mdboom to review the glean MSG changes
@whd to review the script and allowlist updates

Note: The allowlist does need to be updated since it deletes BQ schemas not in it, so although MGS adds them automatically, they still have to be included (I was wrong earlier).

### DIFF
--- a/allowlist
+++ b/allowlist
@@ -1,6 +1,8 @@
 telemetry/core
 org-mozilla-fenix/.*
 org-mozilla-reference-browser/.*
+org-mozilla-samples-glean/.*
+org-mozilla-tv-firefox/.*
 activity-stream/.*
 eng-workflow/bmobugs
 mobile/activation
--- a/bin/schema_generator.sh
+++ b/bin/schema_generator.sh
@@ -124,7 +124,7 @@ function main() {
     clone_and_configure_mps
 
     # Generate new schemas
-    mozilla-schema-generator generate-glean-ping --out-dir . --pretty
+    mozilla-schema-generator generate-glean-pings --out-dir .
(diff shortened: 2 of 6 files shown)


allowlist
```diff
@@ -1,6 +1,8 @@
 telemetry/core
 org-mozilla-fenix/.*
 org-mozilla-reference-browser/.*
+org-mozilla-samples-glean/.*
+org-mozilla-tv-firefox/.*
 activity-stream/.*
```
whd: It is my understanding that `glean/.*` is not intended to be added to the allowlist, is that correct?

bin/schema_generator.sh
```diff
@@ -23,7 +23,7 @@ set -exuo pipefail
 MPS_REPO_URL="git@github.com:mozilla-services/mozilla-pipeline-schemas.git"
 MPS_BRANCH_SOURCE="dev"
 MPS_BRANCH_WORKING="local-working-branch"
-MPS_BRANCH_PUBLISH="generated-schemas"
+MPS_BRANCH_PUBLISH="test-generated-schemas"
```
whd:
```suggestion
MPS_BRANCH_PUBLISH="generated-schemas"
```

allowlist
fbertsch: Right. Glean is a library that sends no telemetry on its own.

bin/schema_generator.sh
fbertsch: Good catch. I'll open an issue about making this configurable for dev testing.

PR #29, opened 2019-06-06
Add non-sync non-main non-pioneer telemetry doctypes to allowlist

whd:
It looks like the schema rollouts so far have been successful. Bigquery tables are case-sensitive so currently e.g. `untrustedModules` will be used as-is in the bq table name, though we could normalize it if we wanted to.


fbertsch:
We could consider moving to a disallow list, after this merge.

whd:
+1 to moving to dissallowlist after this merge.

PR #32, opened 2019-06-12
Remove mobile-metrics from the allowlist

whd:
See https://github.com/mozilla/gcp-ingestion/issues/649 for details


whd:
It indeed appears to be a transpiler issue. @jklukas is following up on that.

Issue #28, opened 2019-06-05
Add dev testing option for schema update script


fbertsch:
Right now, the only option for the schema update script testing is to either:
- Use the prod credentials and push to mozilla/generated-schemas (bad!)
- Change the script to push to a different repo or branch, and use a personal ssh key

We should allow the user to set their own repo, e.g.`fbertsch/mozilla-pipeline-schemas` instead of `mozilla/mozilla-pipeline-schemas`. Then they can use their own personal ssh key, and production can set that to `mozilla` with the prod ssh key.

PR #37, opened 2019-06-25
Expose a public interface to the schema generator script for testing

acmiyaguchi:
This exposes `MPS_REPO_URL`, `MPS_BRANCH_SOURCE`, and `MPS_BRANCH_PUBLISH` as a public interface to the container's main entry point. This also defaults the publish branch to `test-generated-schemas` when running the job locally. This prevents accidentally pushing to the production branch when testing locally. 

Fixes https://github.com/mozilla/mozilla-schema-generator/issues/28.

### DIFF
--- a/bin/schema_generator.sh
+++ b/bin/schema_generator.sh
@@ -8,10 +8,17 @@
 # Environment variables:
 #   MPS_SSH_KEY_BASE64: A base64-encoded ssh secret key with permissions to push
 #                       to mozilla-pipeline-schemas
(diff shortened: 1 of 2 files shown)


acmiyaguchi:
It's actually set to test-generated-schemas by default when it's run via `make run` through `docker-compose`. This doesn't help when people run the container from dockerhub directly though. 

We can push this through and set the environment variable in airflow first, then come back and set the defaults to something reasonable.

PR #39, opened 2019-06-28
Represent histograms as strings

### LINKED ISSUE #38: Represent histograms as strings
We have agreement to move forward with representing histograms as json string columns and we need to make the relevant changes here.

Keyed histograms should end up as key/value structs.

jklukas:
Fixes #38

### DIFF
--- a/mozilla_schema_generator/probes.py
+++ b/mozilla_schema_generator/probes.py
@@ -51,39 +51,8 @@ class MainProbe(Probe):
 
     first_added_key = "first_added"
 
-    # TODO: Do we use log_sum_*, sum_squares_*?
     histogram_schema = {
-        "type": "object",
-        "properties": {
-            "bucket_count": {
-                "minimum": 0,
-                "type": "integer"
-            },
-            "histogram_type": {
-                "minimum": 0,
-                "type": "integer"
-            },
-            "range": {
-                "items": {
-                    "type": "integer"
-                },
-                "type": "array"
-            },
-            "sum": {
-                "minimum": 0,
-                "type": "integer"
-            },
-            "values": {
-                "additionalProperties": False,
-                "patternProperties": {
-                    "^[0-9]+$": {
-                        "minimum": 0,
-                        "type": "integer"
-                    }
-                },
-                "type": "object"
-            }
-        }
+        "type": "string",
     }
 
     def __init__(self, identifier: str, definition: dict):
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -213,19 +213,24 @@ class TestIntegration(object):
 
     def test_max_size(self, schema, env, probes):  # noqa F811
         # Test that we split into multiple tables when we exceed max_size of columns
-        probes["histogram/third_probe"] = {
-            "history": [
-                {
-                    "second_level": False,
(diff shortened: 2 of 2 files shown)


mozilla_schema_generator/probes.py
```diff
@@ -51,39 +51,8 @@ class MainProbe(Probe):
...
-                "patternProperties": {
-                    "^[0-9]+$": {
-                        "minimum": 0,
-                        "type": "integer"
-                    }
-                },
-                "type": "object"
-            }
-        }
+        "type": "string",
```
jklukas: This appears to be the only change needed in the actual code here?

In order to verify this made the BQ schema I wanted, I had to pipe the json schema output through jq:

```
mozilla-schema-generator generate-main-ping | jq -r '.full[0]' > histo_string.json
jsonschema-transpiler -t bigquery histo_string.json > histo_string.bq
```

I assume that the bin/schema_generator.sh machinery handles parsing out the top level in the produced `.json` file?

jklukas: There is one failing test that I haven't been able to modify successfully, as I'm not yet familiar enough with the plumbing here. In `test_integration.py`, `test_max_size` fails to split. I tried reducing max_size down to 6, but it still doesn't split. If I reduce further, I get an error about max_size being smaller than env. Should I add additional probes in order to increase the column count on this test?

jklukas:
cc @acmiyaguchi @fbertsch

mozilla_schema_generator/probes.py
acmiyaguchi: This should be the only modification that's needed, I'm not sure about the test though.

It sounds like you have an out of date version of jsonschema-transpiler, as of v1.0, the BigQuery schemas [return the root as a list](https://github.com/mozilla/jsonschema-transpiler/releases/tag/1.0.0). You can upgrade using `cargo install jsonschema-transpiler --force`. It also reads directly from stdin, so you can pipe everything in a single line.

jklukas: I'm on jsonschema-transpiler 1.0 and it's doing what I expect; it's m-s-g that's putting out a json file that starts with `{"full": [ ...` that I'm not expecting.

acmiyaguchi: Interesting, I don't think we're using the generated main ping schema yet. It looks like the schema_generator script needs to be modified accordingly to get it in the daily updates. 

Maybe around or before here?

https://github.com/mozilla/mozilla-schema-generator/blob/a24083c9b77a77d949ba859c9eb7db2e6b8ea7cc/bin/schema_generator.sh#L108-L110

jklukas:
With this representation, I'm getting a column count of 4909 for main ping.

fbertsch:
We still need to change this to output `main/main.1.schema.json` instead of `full/full.1.schema.json`

fbertsch:
I suppose I can address those in a followup PR, since this doesn't actually add the generated main schemas to MPS

jklukas:
> I suppose I can address those in a followup PR, since this doesn't actually add the generated main schemas to MPS

I would love that. Will create an issue. Are we good to merge this, then?

fbertsch:
Yup, let's merge!

Issue #43, opened 2019-07-05
Remove userPrefs from main ping schema


jklukas:
Per @fbertsch, we will end up with a forwards-incompatible BQ table if allow userPrefs to become fields.

PR #44, opened 2019-07-08
Add main ping

fbertsch:
Changes available here: https://github.com/mozilla-services/mozilla-pipeline-schemas/tree/test-generated-schemas

Closes https://github.com/mozilla/mozilla-schema-generator/issues/43


mozilla_schema_generator/__main__.py
```diff
@@ -144,7 +144,7 @@ def write_schema(repo, repo_id, config, out_dir, split, pretty, generic_schema):
     dump_schema(schemas, out_dir.joinpath(repo_id), pretty)
 
 
-def dump_schema(schemas, out_dir, pretty):
+def dump_schema(schemas, out_dir, pretty, *, version=1):
```
acmiyaguchi: I haven't seen `*` used like this before, this is the same pattern as `*args, **kwargs`? Why not omit the `*`?

mozilla_schema_generator/main_ping.py
```diff
@@ -8,64 +8,25 @@
...
-        schema._delete_key(("properties", "payload", "properties", "log"))
-
-        # 6. Update slices
-        for p in ("parent", "content", "gpu"):
-            schema.set_schema_elem(("properties", "payload", "properties", "processes", "properties", p, "properties", "gc", "properties", "random", "items", "properties", "slices", "type"), "number") # noqa E501
-            schema.set_schema_elem(("properties", "payload", "properties", "processes", "properties", p, "properties", "gc", "properties", "worst", "items", "properties", "slices", "type"), "number") # noqa E501
-
-        # 7. Ignore threadHangStats, deprecated in 57
-        schema._delete_key(("properties", "payload", "properties", "threadHangStats"))
-
```
acmiyaguchi: Nice.

mozilla_schema_generator/main_ping.py
```diff
@@ -8,64 +8,25 @@
...
         return self._update_env(schema)
 
     def _update_env(self, schema):
-        # 2. Make partnerNames just a str array
-        schema.set_schema_elem(("properties", "environment", "properties", "partner", "properties", "partnerNames", "type"), "array") # noqa E501
-        schema.set_schema_elem(("properties", "environment", "properties", "partner", "properties", "partnerNames", "items"), {"type": "string"}) # noqa E501
+        schema._delete_key(prepend_properties(("environment", "settings", "userPrefs")))
```
acmiyaguchi: It's worth adding a note about why we're not adding `userPrefs` to the schema.

mozilla_schema_generator/__main__.py
robhudson: This is new to me too so I looked it up. [PEP 3102](https://www.python.org/dev/peps/pep-3102/) describes it in detail. Basically it forces anything after the `*` to be a keyword parameter and doesn't allow any more positional args aside from the ones defined like `*args` would do.

mozilla_schema_generator/main_ping.py
jklukas: Does this line address #43?

acmiyaguchi: It does. Cases 4, 5, and 7 in the previous revision use the `_delete_key` method and the `prepend_properties` function just interleaves every other element in the list with a "property" key for traversing the JSON Schema.

PR #45, opened 2019-07-09
Add remaining telemetry pings

whd:
Requires https://github.com/mozilla-services/cloudops-infra/pull/1178

We may need to back this out again if the additional pings cause issues, but we're not expecting that to be the case.

Sync pings are still an open question but can hopefully be dealt with in a backwards-compatible manner from the [current placeholder schema](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/schemas/telemetry/sync/sync.4.schema.json).


whd:
The bq sink dataflow job has been redeployed using the schemas artifact so this should be ready to merge.

PR #41, opened 2019-07-02
Drop under-specified columns during transpilation

fbertsch:
Commit with differences here: https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/ae10d1e4d8efa88aaaf80a15a29b35d71533a515

Looks like just a few schemas will change (and the prio ping added?)

PR #46, opened 2019-07-16
Snake case BQ table and field names

jklukas:
Relies on https://github.com/mozilla/jsonschema-transpiler/pull/79

Closes #41


Dockerfile
```diff
@@ -24,7 +24,7 @@ ENV CARGO_INSTALL_ROOT=${HOME}/.cargo
 ENV PATH ${PATH}:${HOME}/.cargo/bin
 
 # Install a tagged version of jsonschema-transpiler
-RUN cargo install jsonschema-transpiler --version 1.0.0
+RUN cargo install jsonschema-transpiler --version 1.2.0
```
jklukas: I'm assuming the jsonschema-transpiler changes will be deployed as 1.2.0

bin/schema_generator.sh
```diff
@@ -146,7 +146,9 @@ function main() {
     # Add transpiled BQ schemas
     find . -type f -name "*.schema.json" | while read -r fname; do
         bq_out=${fname/schema.json/bq}
-        jsonschema-transpiler --type bigquery "$fname" > "$bq_out"
+        # Snake case untrustedModules for bug 1565074
+        bq_out=${bqout/untrustedModules/untrusted_modules}
```
whd: s/bqout/bq_out/

bin/schema_generator.sh
```diff
@@ -116,7 +116,7 @@ function commit_schemas() {
     # Keep only the schemas dir
     find . -mindepth 1 -maxdepth 1 -not -name .git -exec rm -rf {} +
     git checkout $MPS_BRANCH_WORKING -- schemas
-    git commit -a -m "Auto-push from schema generation" || echo "Nothing to commit"
+    git commit -a -m "Auto-push from schema generation [ci skip]" || echo "Nothing to commit"
```
jklukas: Closes https://github.com/mozilla-services/mozilla-pipeline-schemas/issues/337

bin/schema_generator.sh
```diff
@@ -146,7 +146,10 @@ function main() {
     # Add transpiled BQ schemas
     find . -type f -name "*.schema.json" | while read -r fname; do
         bq_out=${fname/schema.json/bq}
-        jsonschema-transpiler --type bigquery "$fname" > "$bq_out"
+        # Snake case untrustedModules for bug 1565074
+        bq_out=${bq_out//untrustedModules/untrusted_modules}
```
whd: While this will work, the double slash has a [different meaning](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) and is probably a typo.

jklukas: My intention here is that double slash means replace all:

> If pattern begins with ‘/’, all matches of pattern are replaced with string. Normally only the first match is replaced

whd: Ok this actually brings up an interesting point that this new directory may not exist since we're replacing `untrustedModules` multiple times. Likely we'll need to do something about that.

jklukas: `bq_out` is a path like `telemetry/untrustedModules/untrustedModules.1.bq` and we need to replace both those instances of `untrustedModules` in that string.

whd: See edit above about the directory not existing. I think that for the purposes of downstream if only the second part is changed we're good, but I would have to check when I get back. It's a little awkward to have the json schema in untrustedModules and the bq schema in a different directory but as long as it works I'm ok with it.

jklukas: > Ok this actually brings up an interesting point that this new directory may not exist since we're replacing untrustedModules multiple times. Likely we'll need to do something about that.

In the last force-push, I added a mkdir call on the next line that should handle that.

> additionally we use by convention - in docTypes, so changing this to _ may have effects if the downstream jenkins script is assuming the directory name and first part of the file name match

Good point. This should be `bq_out=${bq_out//untrustedModules/untrusted-modules}`. I think if I make that change, we should be good to go without changing the jenkins logic. We aren't doing the full snake casing here; we're only handling untrustedModules as a special case.

Dockerfile
acmiyaguchi: Yes, the package has been [tagged](https://github.com/mozilla/jsonschema-transpiler/releases/tag/v1.2.0) and [deployed to cargo](https://crates.io/crates/jsonschema-transpiler). It's ready for testing and use.

PR #49, opened 2019-08-01
Add disableSHA1rollout to specially snake-cased pings

whd:


bin/schema_generator.sh
```diff
@@ -144,9 +144,10 @@ function main() {
     # Add transpiled BQ schemas
     find . -type f -name "*.schema.json" | while read -r fname; do
         bq_out=${fname/schema.json/bq}
-        # Normalize name of untrustedModules ping for bug 1565074;
+        # Normalize names of pings for bug 1565074;
         # the // means "replace all".
         bq_out=${bq_out//untrustedModules/untrusted-modules}
+        bq_out=${bq_out//disableSHA1rollout/disable-sha1rollout}
```
jklukas: I would have expected this to be `disable-sha1-rollout`. Are we sure this is how it's getting transformed?

whd: Here's one such error message: gs://moz-fx-data-stage-data/telemetry-decoded_bq-sink-raw/error/2019-08-01/03/error-2019-08-01T03-10-00.000Z-2019-08-01T03-20-00.000Z-0-00001-of-00060.ndjson.gz

jklukas: I checked with the Java implementation and indeed there's only the one break.

PR #54, opened 2019-08-12
Enable all_pings for Glean

fbertsch:
- Allows `all_pings` for a Glean probe definition, where
  the metric is added to every ping for that app
- Correctly sorts the Glean probe definitions by date
- Move sorting of Main probes to the MainProbe class
- Update tests to fix new schema for Main probe defs
  (it now matches the probe-info-service exactly)


fbertsch:
@mdboom ah, I left a file out of my commit! Thanks for letting me know. This one adds tests for those probe changes.

PR #58, opened 2019-08-15
Add support for schema aliasing

### LINKED ISSUE #53: Generate the saved-session ping schema
This should mirror the `main` ping, but with a different `docType`.

We could even consider simply copying the BQ schema file over if it is an exact match.

fbertsch:
Resolves #53, for both saved-session and first-shutdown.

This adds a new file which lists the schemas we want to alias.
This does not handle versions; the aliased namespace/doctype
are assumed to have the same versions as the source schemas.

This also includes a validation schema for the aliasing file,
and tests that the file is valid.

The test output is here: https://github.com/mozilla-services/mozilla-pipeline-schemas/tree/test-generated-schemas

### DIFF
--- a/aliases.json
+++ b/aliases.json
@@ -1,12 +1,12 @@
 {
   "telemetry": {
     "saved-session": {
-      "src-namespace": "telemetry",
-      "src-doctype": "main"
+      "source-namespace": "telemetry",
+      "source-doctype": "main"
     },
     "first-shutdown": {
-      "src-namespace": "telemetry",
-      "src-doctype": "main"
+      "source-namespace": "telemetry",
+      "source-doctype": "main"
     }
   }
 }
--- a/aliases.json
+++ b/aliases.json
@@ -1,0 +1,12 @@
+{
+  "telemetry": {
+    "saved-session": {
+      "src-namespace": "telemetry",
+      "src-doctype": "main"
+    },
+    "first-shutdown": {
+      "src-namespace": "telemetry",
+      "src-doctype": "main"
+    }
+  }
+}
--- a/bin/alias_schemas
+++ b/bin/alias_schemas
@@ -57,7 +57,7 @@ def main(aliases_path, base_dir):
                 fname = source.name.replace(source_doctype, dest_doctype)
                 dest = dest_dir / fname
 
-                print("Aliasing {} to {}".format(dest, source))
+                print(f"Aliasing {dest} to {source}")
 
                 dest.parent.mkdir(parents=True, exist_ok=True)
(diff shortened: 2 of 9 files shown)


fbertsch:
One change we might want to make is to move this _before_ `filter_schemas` is run, so that we can alias schemas we're filtering out and/or filter out schemas that are aliased (if temporary).

aliases.json
```diff
@@ -0,0 +1,12 @@
+{
+  "telemetry": {
+    "saved-session": {
+      "src-namespace": "telemetry",
+      "src-doctype": "main"
```
acmiyaguchi: It might be worth either dropping the `src-` prefix (simply `doctype`) or to avoid the abbreviation e.g. `source-doctype`.

bin/alias_schemas
```diff
@@ -0,0 +1,51 @@
...
+def main(aliases_path, base_dir):
+    base_dir = pathlib.Path(base_dir)
+
+    with open(aliases_path) as f:
+        aliases_parsed = json.load(f)
+
+    for dest_ns, doctypes in aliases_parsed.items():
+        for dest_dt, alias_info in doctypes.items():
+            source_ns = alias_info['src-namespace']
+            source_dt = alias_info['src-doctype']
```
acmiyaguchi: The naming is all over the place, I'd consider making the names consistent. For example:

```
src-namespace -> source-namespace
source_ns -> source_namespace
```

bin/alias_schemas
```diff
@@ -0,0 +1,51 @@
...
+            source_dir = base_dir / source_ns / source_dt
+            dest_dir = base_dir / dest_ns / dest_dt 
+            for source in source_dir.glob('*'):
+                fname = source.name.replace(source_dt, dest_dt)
+                dest = dest_dir / fname
+
+                print("Aliasing {} to {}".format(dest, source))
```
acmiyaguchi: f-strings are awesome, if you're willing to stick to python3.6+.

```suggestion
                print(f"Aliasing {dest} to {source}")
```

fbertsch: Haven't used them yet, thanks for the info!

fbertsch:
@acmiyaguchi RFAL

fbertsch:
@whd the reason we didn't go that path is we would still need to re-generate the schemas for every one. Instead we'd like to be able to alias on the _generated_ schemas as well, so it had to happen after that step.

whd:
Another note, it's potentially worth adding some documentation to the top-level README with this functionality.

PR #59, opened 2019-08-15
Move jsonschema requirement

whd:
I re-ran MSG from airflow and it looks like only test requirements were updated for jsonschema. See [the log](https://workflow.telemetry.mozilla.org/log?task_id=mozilla_schema_generator&dag_id=probe_scraper&execution_date=2019-08-14T00%3A00%3A00%2B00%3A00).


whd:
This appears to still be failing, but perhaps the container wasn't updated when I re-ran the job.

PR #60, opened 2019-08-16
Requirements fail

fbertsch:
Before, the requirements weren't being installed during the image creation - now they are. This should guarantee that the script has access to them.


fbertsch:
Okay, moved all requirements installs to the build step. A test run is here: https://github.com/fbertsch/mozilla-pipeline-schemas/tree/test-generated-schemas/

fbertsch:
@acmiyaguchi it is, because the install was happening in a virtualenv, the script didn't have access to them.

PR #64, opened 2019-08-19
Ignore camelCase docTypes

jklukas:
Companion to https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/391

### DIFF
--- a/bin/schema_generator.sh
+++ b/bin/schema_generator.sh
@@ -133,11 +133,12 @@ function main() {
         if [[ $fname =~ metadata/sources ]] ; then
             continue
         fi
+        # This is a legacy camelCase docType and should be ignored in favor of
+        # its kebab-case replacement.
+        if [[ $fname =~ [^-_.a-z0-9] ]] ; then
+            continue
+        fi
         bq_out=${fname/schema.json/bq}
-        # Normalize names of pings for bug 1565074;
-        # the // means "replace all".
-        bq_out=${bq_out//untrustedModules/untrusted-modules}
-        bq_out=${bq_out//disableSHA1rollout/disable-sha1rollout}
         mkdir -p $(dirname "$bq_out")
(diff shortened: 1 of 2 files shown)


fbertsch:
@jklukas what do you think about using the disallow list for this instead?

bin/schema_generator.sh
```diff
@@ -133,11 +133,12 @@ function main() {
         if [[ $fname =~ metadata/sources ]] ; then
             continue
         fi
+        # This is a legacy camelCase docType and should be ignored in favor of
+        # its kebab-case replacement.
+        if [[ $fname =~ [^-_.a-z0-9] ]] ; then
```
whd: Is there a reason not to use `[A-Z]` here instead? This special processing is specific to camelCase and the regex encodes the snake_case special case since that doesn't require said processing.

jklukas:
> @jklukas what do you think about using the disallow list for this instead?

Considering we have only two specific cases to handle, I like the idea of using the disallow list. Will change.

bin/schema_generator.sh
```diff
@@ -134,10 +134,6 @@ function main() {
             continue
         fi
```
jklukas: `metadata/sources` is in the `disallowlist`, so I expect that would exclude it. Is this case still needed?

I think we should probably add a `metadata/.*` line to `disallowlist` and remove all the specific cases, as MPS itself documents that the `metadata` directory is not a real namespace, so I never expect us to add a real ping type there.

whd: What the disallowlist does is prevent checking in schemas on commit. What this special case does is prevent BQ schema generation in the first place, since processing this AWS-specific schema results in an exception.

Metadata contains the schemas for the decoded, raw, and error tables, which are special cases handled by the ops logic separately.

fbertsch: @whd this question is up your alley

PR #71, opened 2019-09-26
Add --tuple-struct to schema generation

whd:
My understanding is that this can roll out independently from https://github.com/mozilla/gcp-ingestion/pull/847. The probe scraper dag is running now so perhaps we should merge+deploy this tomorrow.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ ENV CARGO_INSTALL_ROOT=${HOME}/.cargo
 ENV PATH ${PATH}:${HOME}/.cargo/bin
 
 # Install a tagged version of jsonschema-transpiler
-RUN cargo install jsonschema-transpiler --version 1.4.1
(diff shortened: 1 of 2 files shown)


acmiyaguchi:
Here is the final diff that would be a result of merging this: https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/d93c6040e18e7a3e7207a8ce509f96eb781b11fe

acmiyaguchi:
I bumped the transpiler to v1.6.0 to fix a bug related to nested arrays (which occur in the untrusted-modules and third-party-modules pings). The new testing schemas are found here: https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/c6b8a0339057deb75ba9befacd966271f9bf6721

acmiyaguchi:
Also, this should not be merged in until https://github.com/mozilla/gcp-ingestion/pull/847 is completed and tested.

Issue #72, opened 2019-09-26
Deploying `mozilla-pipeline-schemas` seems to be missing changes


mdboom:
In the process of investigating whether [this PR](https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/416) had been deployed, I looked at the history of the generated-schemas branch, and looking at [this commit](https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/4d1226a54ec1c8b6c157bb76ec79c0e5e9a46115#diff-bbe24176de7c6e74422eae5a6698e520) noticed that the changes were not applied to all of the schemas that are derived from the glean schema.

The removed of the `pattern` regex seems to only appear on the `glean` schema, not any of the `org-mozilla-*` schemas (which are derived from the same template, so that's surprising).

fbertsch:
It looks like we are actually [getting the glean schema from the dev branch](https://github.com/mozilla/mozilla-schema-generator/blob/master/mozilla_schema_generator/glean_ping.py#L23). Testing now to see if switching to master gives the desired effect.

fbertsch:
Fixed with using master branch. Effect was https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/4cf74286f012912b599b2e65272c7fdf16e20f00.

PR #73, opened 2019-09-26
Use master instead of dev for Glean

fbertsch:
Results from a test here: https://github.com/mozilla-services/mozilla-pipeline-schemas/tree/frank-test-generated-schemas


fbertsch:
Fixes #72

fbertsch:
@mdboom I also just realized we need to update the config here when we add new metrics. Let's keep that in mind as we keep adding new ones.

PR #80, opened 2019-11-22
Bug 1595899 - Add XPI field to main ping schema

relud:


acmiyaguchi:
Pulling in @fbertsch for a second opinion on the extra changes.

fbertsch:
@acmiyaguchi and I chatted about this last week (apologies for not responding earlier). It looks like this is an unexpected change for those other schemas. We should try to only add the single XPI field, if possible.

relud:
I've updated this to only add the single XPI field, resulting in mozilla-services/mozilla-pipeline-schemas@14eba8d

I'm unclear on how this impacted these schemas in addition to `main.4.bq`, but did still only add the single XPI field:

    schemas/telemetry/first-shutdown/first-shutdown.4.bq
    schemas/telemetry/saved-session/saved-session.4.bq

acmiyaguchi:
`first-shutdown` and `saved-session` are copies of the main ping schema, so they should mirror the behavior in the main ping.

relud:
Upon further inspection this change is unnecessary.

This was implemented in hopes of removing references to `additional_properties` from `main_summary_v4`, but that still won't happen because `payload.processes.dynamic.scalars` and `payload.processes.dynamic.keyedScalars` are still needed and can't be added the to schema.

PR #82, opened 2019-12-12
Fetch ping names from ping definition list

### LINKED ISSUE #81: Read Glean pings from the `pings` endpoint
Previously, we retrieved pings from the metrics file. We now have a separate endpoint that lists all the pings explicitly. We should use that instead to get the list.

It should be updating this piece of code: https://github.com/mozilla/mozilla-schema-generator/blob/master/mozilla_schema_generator/glean_ping.py#L91

badboy:
It might be as easy as this, we do need to:

* Land the probe-scraper changes first
* Then add the pings.yaml files to the probe-scraper correctly for all projects

Fixes #81

### DIFF
--- a/mozilla_schema_generator/glean_ping.py
+++ b/mozilla_schema_generator/glean_ping.py
@@ -72,29 +72,29 @@ class GleanPing(GenericPing):
         logging.info(f"For {self.repo}, found Glean dependencies: {dependencies}")
         return dependencies
 
-    def _get_probe_defn_list(self) -> List[Dict]:
-        probes = self._get_json(self.probes_url)
-        items = list(probes.items())
+    def get_probes(self) -> List[GleanProbe]:
+        data = self._get_json(self.probes_url)
+        probes = list(data.items())
         for dependency in self.get_dependencies():
             dependency_probes = self._get_json(
                 self.probes_url_template.format(dependency)
             )
-            items += list(dependency_probes.items())
+            probes += list(dependency_probes.items())
 
-        return items
-
-    def get_probes(self) -> List[GleanProbe]:
-        probes = self._get_probe_defn_list()
         pings = self.get_pings()
 
         return [GleanProbe(_id, defn, pings=pings) for _id, defn in probes]
 
     def get_pings(self) -> Set[str]:
-        url = self.ping_url_template.format(self.repo)
-        pings = GleanPing._get_json(url)
-        addl_pings = set(pings.keys())
+        try:
+            url = self.ping_url_template.format(self.repo)
+            pings = GleanPing._get_json(url)
+            addl_pings = set(pings.keys())
 
-        return self.default_pings | addl_pings
+            return self.default_pings | addl_pings
+        except HTTPError:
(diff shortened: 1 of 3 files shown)


mozilla_schema_generator/glean_ping.py
```diff
@@ -89,12 +90,9 @@ def get_probes(self) -> List[GleanProbe]:
         return [GleanProbe(_id, defn, pings=pings) for _id, defn in probes]
 
     def get_pings(self) -> Set[str]:
-        probes = [GleanProbe(_id, defn) for _id, defn in self._get_probe_defn_list()]
```
fbertsch: Can you also remove the `_get_probe_defn_list` function, since `get_pings` was the only reason it was pulled out as a separate piece? It can be consolidated into `get_probes`.

fbertsch: Do _not_ do the above if we proceed with the fallback.

badboy: I don't think we need a fallback. If there's no extra pings.yaml it's only the default pings anyway

badboy:
For me tests hang locally so I can't actually verify that it works correctly :/

badboy:
hanging tests: local network problem. I can run them now.

Now they should actually pass

mozilla_schema_generator/glean_ping.py
```diff
@@ -71,32 +72,29 @@ def get_dependencies(self):
...
-
-        return self.default_pings | addl_pings
+        try:
+            url = self.ping_url_template.format(self.repo)
+            pings = GleanPing._get_json(url)
+            addl_pings = set(pings.keys())
+
+            return self.default_pings | addl_pings
+        except HTTPError:
+            # If we get an error (e.g. 404 Not Found) we fall back to only the default pings
```
fbertsch: 👍

mozilla_schema_generator/glean_ping.py
```diff
@@ -22,11 +22,12 @@ class GleanPing(GenericPing):
 
     schema_url = "https://raw.githubusercontent.com/mozilla-services/mozilla-pipeline-schemas/master/schemas/glean/glean/glean.1.schema.json" # noqa E501
     probes_url_template = "https://probeinfo.telemetry.mozilla.org/glean/{}/metrics"
+    ping_url_template = "https://probeinfo.telemetry.mozilla.org/glean/{}/pings"
     repos_url = "https://probeinfo.telemetry.mozilla.org/glean/repositories"
     dependencies_url_template = "https://probeinfo.telemetry.mozilla.org/glean/{}/dependencies"
 
     default_dependencies = ['glean']
-    default_pings = {"baseline", "events", "metrics"}
+    default_pings = {"baseline", "events", "metrics", "deletion-request"}
```
fbertsch: `deletion-request` is a new default ping type? 🤔

fbertsch:
@badboy we should get this merged! The tests need to be updated to include the new `deletion-request` default ping (see the CI).

fbertsch:
Fixes #88

badboy:
Checks green, @fbertsch

PR #98, opened 2020-01-17
Add individual commit information into auto-pushed commits

### LINKED ISSUE #20: Auto-pushes do not retain commit information
https://github.com/mozilla-services/mozilla-pipeline-schemas/commits/generated-schemas

The history in the generated-schemas branch is uninformative. It should include the commits/PRs since the last push for auditing purposes.

acmiyaguchi:
This fixes #20 by adding log information of commits between the last generated commit and the present. This works by looking at the last commit date in the generated-schemas branch, and filtering out the commits in the master branch using the `--since` flag. This works best when all commits are added via squash/rebase, so I've disabled merge commits in mozilla-pipeline-schemas. This is easy to revert if undesirable.

The test associated test generates the following branch in mozilla-pipeline-schemas: https://github.com/mozilla-services/mozilla-pipeline-schemas/commits/test-generated-schemas-changelog

```
 Auto-push from schema generation [ci skip]

b696996	2020-01-08 12:51:40 -0500	Make the $id for main ping schema show version as "4"
a56043b	2020-01-08 11:54:56 -0500	Add page for china build to activity stream impression_stats
6b9c323	2020-01-08 10:39:36 -0500	Allow negative sessionLengths and event timestamps
29b042f	2020-01-08 10:26:59 -0500	Allow for old-style ping_type in glean
d093483	2020-01-08 10:26:59 -0500	Add changelog entry
dc87e27	2020-01-08 10:26:59 -0500	Make glean ping_type kebab
f28259d	2020-01-07 15:58:25 -0800	Revert "Add build/ folder to gitignore"
```

There are some quirks around how we handle git management within the container that I think are undesirable. First, every run of `schema_generator.sh` will clobber/delete the state of the previous run, which requires typing in the password for your ssh key to push to the repo. There are also protected files in the checked out repo that will require user prompting in the script. One idea is to prepopulate this repo with a submodule pointing to mozilla-pipeline-schemas, like mozilla-services/edge-validator, or to modifying the `setup_git_ssh` and `clone_and_configure_mps` functions so they can be run without throwing away all the initialization. This would make it easier to generate diffs between versions of the repo or revisions of mps. 

Additionally, we might want to think about adding tags to the master branch that correspond to a generated schema. This could be used as an alternative landmark to commit date for determining the date range for the autocommit comment.

### DIFF
--- a/bin/schema_generator.sh
+++ b/bin/schema_generator.sh
@@ -91,7 +91,7 @@ function filter_schemas() {
     find . -name '*.bq' | grep -f $DISALLOWLIST | xargs rm -f
 }
 
-function commit_schemas() {
+function commit_and_push_schemas() {
     # This method will keep a changelog of releases. If we delete and newly
     # checkout branches everytime, that will contain a changelog of changes.
     # Assumes the current directory is the root of the repository
@@ -110,6 +110,7 @@ function commit_schemas() {
     find . -mindepth 1 -maxdepth 1 -not -name .git -exec rm -rf {} +
     git checkout $MPS_BRANCH_WORKING -- schemas
     git commit -a -m "Auto-push from schema generation [ci skip]" || echo "Nothing to commit"
+    git push
 }
 
 function main() {
@@ -156,8 +157,7 @@ function main() {
 
     # Push to branch of MPS
     cd ../
-    commit_schemas
-    git push
+    commit_and_push_schemas
 }
 
 main "$@"
--- a/bin/schema_generator.sh
+++ b/bin/schema_generator.sh
@@ -16,11 +16,6 @@
 # Example usage:
 #   export MPS_SSH_KEY_BASE64=$(cat ~/.ssh/id_rsa | base64)
 #   make build && make run
-#
-# TODO: Update schema mapping for validation
-# TODO: Handle overwriting glean schemas
-# TODO: Include Main Ping from schema generation
-# TODO: What the heck to do with pioneer-study, a non-nested namespace
 
 set -exuo pipefail
 
--- a/bin/schema_generator.sh
+++ b/bin/schema_generator.sh
@@ -106,7 +106,7 @@ function create_changelog() {
         --date=iso
 }
 
-function commit_and_push_schemas() {
+function commit_schemas() {
     # This method will keep a changelog of releases. If we delete and newly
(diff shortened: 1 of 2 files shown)


bin/schema_generator.sh
```diff
@@ -87,7 +86,28 @@ function filter_schemas() {
...
+    # isn't enforced, it's possible to miss out on certain commits to the
+    # generated branch. Another solution is to generate a tag on master before
+    # the generator is run. However, the heuristic of using the latest commit
+    # date works well enough.
+    start_date=$(git log "${MPS_BRANCH_PUBLISH}" -1 --format=%cd --date=iso)
+    git log "${MPS_BRANCH_SOURCE}" \
+        --since="$start_date" \
+        --pretty=format:"%h%x09%cd%x09%s" \
+        --date=iso \
+        | cat
```
acmiyaguchi: `git log` writes to a pager by default (e.g. less), so piping to cat allows this to be used as input elsewhere.

whd: git has a `--no-pager` flag as well that you should probably leverage

acmiyaguchi: After playing around with the option, it looks like there wasn't actually a problem with paging. I reran the script without the pipe into cat, and it looks to be working correctly.

tests/test_commit_changelog.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+# A script for testing the functionality of changelog information in generated
+# schemas. The output of the this command must be checked manually, and this
```
whd:
```suggestion
# schemas. The output of this command must be checked manually, and this
```

bin/schema_generator.sh
```diff
@@ -55,6 +50,10 @@ function setup_git_ssh() {
     chown -R "$(id -u):$(id -g)" "$HOME/.ssh"
     chmod 700 "$HOME/.ssh"
     chmod 700 "$HOME/.ssh/id_ed25519"
+
+    # add private key to the ssh agent to prompt for password once
```
whd: I'm assuming this is a noop on CI in that the ssh key doesn't have a passphrase.

acmiyaguchi: This should be the case. This also only runs for the duration of the script session.

PR #99, opened 2020-01-17
Use timestamp for latest probes file

fbertsch:


mozilla_schema_generator/generic_ping.py
```diff
@@ -150,5 +150,5 @@ def _get_json(url: str) -> dict:
             # For probe-info-service requests, add
             # random query param to force cloudfront
             # to bypass the cache
-            url += f'?r={random.randint(0,100)}'
+            url += f'?t={datetime.datetime.utcnow().isoformat()}'
```
robhudson: I was going to suggest epoch time `int(time.time())` to avoid special characters in the time string with the ISO time format, but either should work to bust cache.

PR #103, opened 2020-02-12
Bug 1597782 -Collect GPU process histograms correctly

tdsmith:
`payload.processes.gpu.histograms` should collect unkeyed histograms.


tdsmith:
`mozilla-schema-generator generate-main-ping |  jq '.main[0].properties.payload.properties.processes.properties.gpu.properties.histograms.properties.COMPOSITE_TIME'`

Before:
```
null
```

After:
```
{
  "type": "string"
}
```

PR #107, opened 2020-03-03
Bug 1603536 Add descriptions to generated schemas

scholtzan:
This adds support for parsing descriptions from probes and adding them to generated schemas.

### DIFF
--- a/mozilla_schema_generator/probes.py
+++ b/mozilla_schema_generator/probes.py
@@ -19,11 +19,10 @@ class Probe(object):
     name_key = "name"
     history_key = "history"
 
-    def __init__(self, identifier: str, definition: dict, description: str):
+    def __init__(self, identifier: str, definition: dict):
         self.id = identifier
         self.type = definition[self.type_key]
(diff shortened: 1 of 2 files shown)


mozilla_schema_generator/probes.py
```diff
@@ -93,6 +100,12 @@ def _set_dates(self, first_added_value: dict):
         self.first_added = min(vals)
         self.last_change = max(vals)
 
+    def _set_description(self, definition):
+        if "description" in definition:
+            self.description = definition["description"]
```
fbertsch: Looks like this is set both here and in `super().__init__`. I think we can remove the latter and just set it here.

tests/test_probes.py
```diff
@@ -271,6 +271,33 @@ def main_probe_all_defn():
     }
 
 
+@pytest.fixture
+def main_probe_description():
```
fbertsch: Might be simpler to just add this description to the other probe definition.

Issue #110, opened 2020-03-04
Probes should implement __repr__ for repl and debugging


acmiyaguchi:
Running the following snippet:
```
from mozilla_schema_generator import GleanPing
GleanPing("fenix").get_probes()
```

results in the following:

```python
[
<mozilla_schema_generator.probes.GleanProbe object at 0x10e8c3b10>,
<mozilla_schema_generator.probes.GleanProbe object at 0x10e8c3810>, 
<mozilla_schema_generator.probes.GleanProbe object at 0x10cf0bc90>, 
<mozilla_schema_generator.probes.GleanProbe object at 0x10e8d3850>, 
<mozilla_schema_generator.probes.GleanProbe object at 0x10cf83d90>, 
<mozilla_schema_generator.probes.GleanProbe object at 0x10e8cd750>,
...
]
```

I expect a human readable name, such as the name and type, instead of the object reference. Implementing `__repr__` should make this human-friendly.

Issue #85, opened 2019-12-13
Add scalar processing to the deletion-request ping


mreid-moz:
Per [Bug 1602064](https://bugzilla.mozilla.org/show_bug.cgi?id=1602064), the `deletion-request` ping will start reporting scalars. We should support that in the schema generator.

mreid-moz:
@scholtzan There's a related bug @ https://bugzilla.mozilla.org/show_bug.cgi?id=1614416 - if it doesn't add too much work/scope, please consider tackling that one as part of this work.

scholtzan:
Fixed in https://github.com/mozilla/mozilla-schema-generator/pull/120

PR #120, opened 2020-03-10
Bug 1614416 Common ping format implementation

scholtzan:
This PR adds support for propagating changes for schemas outside of the main ping to those that include the common ping format. This fixes issues where columns are missing for some schemas that get added by the schema generator (see https://bugzilla.mozilla.org/show_bug.cgi?id=1614416)

Schemas that follow the common ping format and should be processed are defined in `common_pings.json`. Currently 2 schemas were invalid due to missing types and missing from that list: 
* https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/511/files#diff-4d86bea8f5f8b9760701da04328d9574
* https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/511/files#diff-ed9293105c5e86452f2bbab2ab604106

I filed https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/511 to fix these issues.

This PR also adds scalar processing for the deletion-request ping (see https://github.com/mozilla/mozilla-schema-generator/issues/85). `common_pings.json` allows to specify configs that should be applied when processing schemas.

I also implemented https://github.com/mozilla/mozilla-schema-generator/issues/110 for easier debugging.


mozilla_schema_generator/configs/deletion_request.yaml
```diff
@@ -0,0 +1,13 @@
...
+      scalars:
+        match:
+          table_group: scalars
+          type: scalar
+          details:
+            keyed: false
+            record_in_processes:
+              contains: main
+            record_into_store:
+              contains: deletion-request
```
scholtzan: This is for https://github.com/mozilla/mozilla-schema-generator/issues/85

mozilla_schema_generator/common_ping.py
```diff
@@ -12,23 +12,26 @@
...
     # ONLY DECREMENT, or the schema will change in an incompatible way!
     MIN_FX_VERSION = 30
 
-    schema_url = "https://raw.githubusercontent.com/mozilla-services/mozilla-pipeline-schemas/master/schemas/telemetry/main/main.4.schema.json" # noqa E501
     env_url = "https://raw.githubusercontent.com/mozilla-services/mozilla-pipeline-schemas/master/templates/include/telemetry/environment.1.schema.json" # noqa E501
     probes_url = GenericPing.probe_info_base_url + "/firefox/all/main/all_probes"
 
-    def __init__(self):
-        super().__init__(self.schema_url, self.env_url, self.probes_url)
+    def __init__(self, schema_url):
```
scholtzan: The main difference between `MainPing` and `CommonPing` is that `CommonPing` can process different schemas

mozilla_schema_generator/__main__.py
```diff
@@ -67,6 +74,79 @@ def generate_main_ping(config, out_dir, split, pretty):
...
+
+        m = re.match(SCHEMA_NAME_RE, common_ping["schema_url"])
+        name = m.group(1)
+        version = m.group(2)
+        config = Config(name, config_data)
+
+        if name == "main":
+            schemas = schema_generator.generate_schema(config, split=split)
+        else:
+            # do not split schemas except the main schema
```
fbertsch: At this point we shouldn't ever be splitting the main ping either; we can simply remove that param (it's disabled below).

mozilla_schema_generator/__main__.py
```diff
@@ -67,6 +74,79 @@ def generate_main_ping(config, out_dir, split, pretty):
...
+          "schemas that are outputted. Otherwise "
+          "the schemas will be on one line."),
+)
+@click.option(
+    '--common-pings-config',
+    default="common_pings.json",
+    help=("File containing URLs to schemas and configs "
+          "of pings in the common ping format."),
+)
+def generate_common_pings(config_dir, out_dir, split, pretty, common_pings_config):
```
fbertsch: I don't believe we need `generate_main_ping` any longer.

mozilla_schema_generator/utils.py
```diff
@@ -15,6 +15,9 @@ def _get(_dict: dict, key: Tuple[str]) -> Any:
     Retrieved the nested `key` from a dict
     """
     if key:
+        if key[0] not in _dict:
```
fbertsch: Can you describe why the behavior of `_get` needs to change?

mozilla_schema_generator/probes.py
```diff
@@ -153,15 +158,24 @@ def __init__(self, identifier: str, definition: dict, *, pings: List[str] = None
         super().__init__(identifier, definition)
 
         defn_pings = set(
-            [p for d in definition[self.history_key] for p in d.get("send_in_pings", ["metrics"])]
+            [
```
fbertsch: Did you run black on this? That's fine by me, but in that case it would be good to add it to CI as well.

mozilla_schema_generator/common_ping.py
robhudson: I'm using `MainPing` in the upcoming probe search service to collect the non-glean pings for desktop Firefox. After this change it looks like I would need to provide the full schema URL? The `main.4.schema.json` one?

acmiyaguchi:
I generated a clean diff of this PR last night: https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/a69aa2a7d7d04113fd036d7ce5ca1eb8ed0737ab

One notable change is the appearance of `simple_measurements` in several of the pings, which is likely due to the renaming of the `MainPing` to `CommonPing`.

mozilla_schema_generator/common_ping.py
acmiyaguchi: It might make sense to subclass the MainPing as a type of CommonPing, and to move modifications to `payload.simple_measurements` there. The URL could be fixed there, leaving the API backwards compatible.

scholtzan: I added `MainPing` back as subclass of `CommonPing`. So it should be backwards compatible now. Also `payload.simple_measurements` will now only get set for main ping

mozilla_schema_generator/__main__.py
scholtzan: For backwards compatibility and because main ping is handled slightly different from common pings, I added back `MainPing`. So to generate main ping we'd now still have to call `generate_main_ping`

mozilla_schema_generator/probes.py
scholtzan: I accidentally did, however haven't done so for any other files. I'd say adding black to CI might be best in a followup PR since there'd be a lot of changes.

mozilla_schema_generator/generic_ping.py
```diff
@@ -101,8 +101,11 @@ def make_schemas(env: Schema, probes: List[Probe], config: Config,
         schemas.append(final_schema)
         for s in schemas:
             for key in config.get_match_keys():
-                s.delete_group_from_schema(key + ("propertyNames",), propagate=False)
-                s.delete_group_from_schema(key + ("additionalProperties",), propagate=True)
+                if s.get(key + ("propertyNames",)) is not None:
+                    s.delete_group_from_schema(key + ("propertyNames",), propagate=False)
+
+                if s.get(key + ("additionalProperties",)) is not None:
```
scholtzan: > Can you describe why the behavior of _get needs to change?

Some schemas don't have these fields. Previously, `_get` just returned and exception

mozilla_schema_generator/probes.py
scholtzan: I added a comment in the code. Basically, before `_get` threw exceptions for accessing non-existing fields

mozilla_schema_generator/utils.py
scholtzan: I added a comment in the code. Basically, before `_get` would throw an exception for accessing non-existend fields

mozilla_schema_generator/common_ping.py
robhudson: Great, thank you

scholtzan:
The `simple_measurements` should be resolved now. 

I agree, we should inform people beforehand and see if there are any objections/concerns. I'll send out a note tomorrow. 

I'll check if there are any queries that depend on `additionalProperties`

mozilla_schema_generator/utils.py
fbertsch: Hmm. That was by design, so if an unexpected field was trying to be retrieved, we would know with a hard fail. If it's failing somewhere it sounds like that may be a bug.

mozilla_schema_generator/generic_ping.py
fbertsch: Okay, I see. I'd rather add a flag to `_get` to let it behave differently for these cases, and leave everything else as-is.

scholtzan: I wrapped these 2 cases into `try` ... `except` which might be a little more elegant than adding an extra flag

fbertsch: That sounds like the right thing to do, thanks!

scholtzan:
I sent an email last week and checked for queries that depend on `additional_properties`. I found only one query that was executed (executed last 2 weeks ago) that accessed on of these fields. I notified the creator of the change a few days ago.

So I think deploying these changes should be fine.

scholtzan:
Looking at the diff for for example `voice`, it looks correct to me:

```diff
--- mozilla-pipeline-schemas-generated-schemas/schemas/telemetry/voice/voice.4.bq	2020-03-26 19:23:54.000000000 -0700
+++ mozilla-pipeline-schemas-test-generated-schemas/schemas/telemetry/voice/voice.4.bq	2020-03-30 10:41:08.000000000 -0700
@@ -104,6 +104,11 @@
                     "type": "STRING"
                   },
                   {
+                    "mode": "NULLABLE",
+                    "name": "foreign_install",
+                    "type": "INT64"
+                  },
+                  {
                     "description": "True or false depending on whether the add-on has binary components. This is always false since Firefox 60. This field is only available after the 'sessionstore-windows-restored' topic is notified.",
                     "mode": "NULLABLE",
                     "name": "has_binary_components",
@@ -161,6 +166,16 @@
                     "mode": "NULLABLE",
                     "name": "update_day",
                     "type": "INT64"
+                  },
+                  {
+                    "mode": "NULLABLE",
+                    "name": "user_disabled",
+                    "type": "INT64"
+                  },
+                  {
+                    "mode": "NULLABLE",
+                    "name": "version",
+                    "type": "STRING"
                   }
                 ],
                 "mode": "NULLABLE",
@@ -291,6 +306,11 @@
                 "type": "STRING"
               },
               {
+                "mode": "NULLABLE",
+                "name": "foreign_install",
+                "type": "INT64"
+              },
+              {
                 "description": "True or false depending on whether the theme has binary components. This is always false since Firefox 60.",
                 "mode": "NULLABLE",
                 "name": "has_binary_components",
@@ -804,6 +824,23 @@
             "mode": "NULLABLE",
             "name": "update",
             "type": "RECORD"
+          },
+          {
+            "fields": [
+              {
+                "mode": "NULLABLE",
+                "name": "key",
+                "type": "STRING"
+              },
+              {
+                "mode": "NULLABLE",
+                "name": "value",
+                "type": "STRING"
+              }
+            ],
+            "mode": "REPEATED",
+            "name": "user_prefs",
+            "type": "RECORD"
           }
         ],
         "mode": "NULLABLE",
@@ -1285,7 +1322,6 @@
                 "type": "FLOAT64"
               },
               {
-                "description": "The version of the OS, e.g. '6.1'. This is `null` on failure.",
                 "mode": "NULLABLE",
                 "name": "version",
                 "type": "STRING"
```

PR #122, opened 2020-04-06
[Bug 1614416] Add environment.system.os.hasSuperfetch and environment.system.os.hasPrefetch as common ping columns

scholtzan:
This adds `environment.system.os.hasSuperfetch` and `environment.system.os.hasPrefetch` as common ping columns. 

Closes https://bugzilla.mozilla.org/show_bug.cgi?id=1614416


wlach:
This looks totally reasonable but I'm not all that familiar with this codebase, reassigning to @acmiyaguchi who I see doing related things in the history.

PR #124, opened 2020-04-24
Bug 1611027 - Update activeGMPlugins to be schema compatible via casting

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1611027)

This casts the integer into a boolean, since this is being reported differently in a few pings. https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/489

Diff can be found here: https://github.com/mozilla-services/mozilla-pipeline-schemas/compare/bdfb8fc~2...bdfb8fc

It looks like this is correctly nested.

```
wget https://raw.githubusercontent.com/mozilla-services/mozilla-pipeline-schemas/bdfb8fcab009521b93e386e15c9513519e1852b9/schemas/telemetry/crash/crash.4.bq
cat crash.4.bq | jq -r '.[] | recurse(.fields[]? + {parent: ((.parent? // []) + [.name])}) | .parent + [.name] | join(".")'  | grep active_gm_plugins

environment.addons.active_gm_plugins
environment.addons.active_gm_plugins.key
environment.addons.active_gm_plugins.value
environment.addons.active_gm_plugins.value.apply_background_updates
environment.addons.active_gm_plugins.value.user_disabled
environment.addons.active_gm_plugins.value.version
```

### DIFF
--- a/mozilla_schema_generator/common_ping.py
+++ b/mozilla_schema_generator/common_ping.py
@@ -64,13 +64,8 @@ class CommonPing(GenericPing):
         schema.set_schema_elem(active_addons + ("version",), string)
         schema.set_schema_elem(active_addons + ("userDisabled",), integer)
         schema.set_schema_elem(
-            active_addons
-            + (
-                "activeGMPlugins",
-                "additionalProperties",
-                "properties",
-                "applyBackgroundUpdates",
-            ),
+            prepend_properties(("environment", "addons", "activeGMPlugins"))
+            + ("additionalProperties", "properties", "applyBackgroundUpdates"),
             with_description(
                 boolean,
                 "Cast into a boolean via mozilla-schema-generator. See bug 1611027.",
--- a/mozilla_schema_generator/common_ping.py
+++ b/mozilla_schema_generator/common_ping.py
@@ -43,6 +43,10 @@ class CommonPing(GenericPing):
         }
         boolean = {"type": "boolean"}
 
+        def with_description(dtype: dict, comment: str) -> dict:
+            """Add a description to the types defined above."""
+            return {**dtype, **dict(description=comment)}
+
         active_addons = prepend_properties(("environment", "addons", "activeAddons")) \
             + ("additionalProperties", "properties")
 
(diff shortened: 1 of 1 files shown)


mozilla_schema_generator/common_ping.py
```diff
@@ -59,6 +63,14 @@ def _update_env(self, schema):
         schema.set_schema_elem(active_addons + ("foreignInstall",), integer)
         schema.set_schema_elem(active_addons + ("version",), string)
         schema.set_schema_elem(active_addons + ("userDisabled",), integer)
+        schema.set_schema_elem(
+            prepend_properties(("environment", "addons", "activeGMPlugins"))
+            + ("additionalProperties", "properties", "applyBackgroundUpdates"),
+            with_description(
+                boolean,
```
jklukas, line 70: So far, we support coercion of booleans to integers, but not of integers to booleans. We could change this to an integer field, which would potentially preserve more information if we get strange values that aren't 0 or 1. Otherwise, we'll need to add a case in the live transformation logic to handle casting ints to boolean and capturing such events in a metric.

jklukas: See https://github.com/mozilla/gcp-ingestion/blob/dc628b2ee5cd9300af3d3661ad3727f402afce53/ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/transform/PubsubMessageToObjectNode.java#L554-L564

acmiyaguchi: Reading over my older PR, it looks like we should be casting to an integer anyways. The integer can take [on values outside of [0, 1].](https://sql.telemetry.mozilla.org/queries/70455/source)

acmiyaguchi:
@jklukas I've changed this to cast to an integer instead, the resulting diff can be found here: https://github.com/mozilla-services/mozilla-pipeline-schemas/compare/bdfb8fc~2...a65c243

A few other changes snuck in with the change, but it takes 10ish minutes to generate a clean base.

acmiyaguchi:
I'm going to manually run the schema generator to generate clean diffs that aren't clumped together for the daily redeploy.

PR #126, opened 2020-05-08
Bug 1636317 - Add logic for pioneer-ingestion metadata

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1636317)

This has a dependency on https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/544, but I'm actually not sure what the proper order to land these in are. It shouldn't matter if they're merged in at the same time though.

The following diff was generated using the following command: `make build && MPS_BRANCH_SOURCE=pioneer make run`

https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/6fc366544a62ef805335963f5e62030b9c081537

### DIFF
--- a/bin/schema_generator.sh
+++ b/bin/schema_generator.sh
@@ -73,15 +73,19 @@ function prepare_metadata() {
     local pioneer_metadata="metadata/pioneer-ingestion/pioneer-ingestion.1.schema.json"
     local structured_metadata="metadata/structured-ingestion/structured-ingestion.1.schema.json"
 
-    find ./telemetry \
-        -name "*.schema.json" -type f \
-        -exec metadata_merge $telemetry_metadata {} ";"
-    find . -path "./pioneer-*" \
-        -name "*.schema.json" -type f \
-        -exec metadata_merge $pioneer_metadata {} ";"
-    find . \( -path ./telemetry -o -path ./metadata -o -path "./pioneer-*" \) -prune -o \
-        -name "*.schema.json" -type f \
-        -exec metadata_merge $structured_metadata {} ";"
+    # schema directory structure is enforced by regex at compile-time
+    # shellcheck disable=SC2044
+    for schema in $(find . -name "*.schema.json" -type f); do
+        if [[ "$schema" == "./telemetry/*" ]]; then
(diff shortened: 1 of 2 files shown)


bin/schema_generator.sh
```diff
@@ -70,18 +70,21 @@ function clone_and_configure_mps() {
 
 function prepare_metadata() {
     local telemetry_metadata="metadata/telemetry-ingestion/telemetry-ingestion.1.schema.json"
+    local pioneer_metadata="metadata/pioneer-ingestion/pioneer-ingestion.1.schema.json"
     local structured_metadata="metadata/structured-ingestion/structured-ingestion.1.schema.json"
 
-    find ./telemetry -name "*.schema.json" -type f \
+    find ./telemetry \
+        -name "*.schema.json" -type f \
```
jklukas, line 77: We should consider recasting this logic now as a single loop through schema files with an if/then chain:

```
for schema in $(find . -name "*.schema.json" -type f); do
  if [[ "$schema" ~= "telemetry/.*" ]]; then
    metadata_merge $telemetry_metadata $schema
  elif ...
done
```

That would prevent the need to ensure we match the inverse logic for the structured case.

acmiyaguchi: Apparently relying on the output of find is fragile due the possibility of whitespace and globs in the output (https://github.com/koalaman/shellcheck/wiki/SC2044). The structure of the schema repository is enforced at compile time, so it should be fine.

acmiyaguchi:
I'm not entirely sure why this diff does not simply contain changes between this and the master branch of MSG when pointing at a static branch on MPS: https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/71908fa594f6a3846ff96dcfd360ca3020edb6ca

I will have to debug this further, parallelization of #129 has helped significantly with debugging. In the meanwhile, I've finished up the related decoder changes.

acmiyaguchi:
I've fixed this now, and the diff looks correct. https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/5ae792e6afd65fbd7a6eba90ed1d27b2a95ef820

PR #129, opened 2020-05-11
Fix #128 - Parallelize glean ping generation

### LINKED ISSUE #128: Improve speed of schema generation
It takes 5-10 minutes to generate schemas on my local machine. There are two possible routes to improving performance: caching and parallelization.

If we could prefetch (or cache) all the relevant information for schema generation for glean pings, this would give considerable performance gains. Ideally, the generation process would take no longer than a minute on a local machine. 

```bash
mozilla-schema-generator generate-glean-pings --out-dir . --cache-dir .cache
```
The cache directory would contain the json blobs from the info service, as well as a metadata file for purging the cache after some time.

A second consideration to parallelize the main loop in `generate_glean_pings`. Currently it iterates element by element. Each of the schemas is self-contained in a separate path, so this should be trivial to parallelize.

```python
    for repo_name, repo_id in repos:
        write_schema(repo_name, repo_id, config, out_dir, split, pretty, generic_schema)
```

Finally, it might help to add the schema generation process to CI. It would need to encapsulate the difference between the two revisions of the schema generator to be generally useful. We have observed this as a testing pattern over many PRs. It typically takes me 10-20 minutes to generate a diff if everything goes well on the first try. This does not cover all of the cases -- sometimes there is a depenency between MPS and MSG that requires manual testing, such as changes to the metadata schemas.

acmiyaguchi:
This fixes #128 by using multiprocessing to parallelize glean generation. Generating the glean ping is by far the slowest thing in the process. By default, this will use processes equal to the number of CPUs available. 

Before: `make run  0.45s user 0.20s system 0% cpu 8:23.99 total`
After: `make run  0.40s user 0.15s system 0% cpu 3:27.38 total`

One downside of this is that the logging is interleaved.

### DIFF
--- a/mozilla_schema_generator/__main__.py
+++ b/mozilla_schema_generator/__main__.py
@@ -9,6 +9,7 @@ import sys
 import yaml
 import json
 import re
+import multiprocessing
 from pathlib import Path
 
 from .common_ping import CommonPing
@@ -209,8 +210,11 @@ def generate_glean_pings(config, out_dir, split, pretty, repo, generic_schema):
 
     config = Config("glean", config_data)
 
+    args = []
     for repo_name, repo_id in repos:
-        write_schema(repo_name, repo_id, config, out_dir, split, pretty, generic_schema)
+        args.append((repo_name, repo_id, config, out_dir, split, pretty, generic_schema))
+    with multiprocessing.Pool() as pool:
+        pool.starmap(write_schema, args)
 
 
 def write_schema(repo, repo_id, config, out_dir, split, pretty, generic_schema):


acmiyaguchi:
This will ocassionally hit the following error: `requests.exceptions.HTTPError: 504 Server Error: Gateway Time-out for url: https://probeinfo.telemetry.mozilla.org/glean/glean/pings?t=2020-05-12T00:13:28.140097`. However, this is intermittent.

acmiyaguchi:
Closing in favor of #130, which is an order of magnitude faster.

PR #132, opened 2020-06-09
Fix: Fail schema generation if maximum schema size is exceeded

scholtzan:
There was a bug in the code which resulted in not throwing an error when the maximum schema size gets exceeded. The fix was to check the size of the schema after filling in the probes. 

I also increased the maximum size to 9500 columns.


mozilla_schema_generator/generic_ping.py
```diff
@@ -71,6 +71,9 @@ def generate_schema(self, config: Config, *, split: bool = None, max_size: int =
         if split:
             schemas[self.extra_schema_key] = self.make_extra_schema(schema, probes, configs)
 
+        if any(schema.get_size() > max_size for _, s in schemas.items() for schema in s):
```
scholtzan, line 74: If `split=True` then schemas get split up in a way that it can have `max_size` columns. That's why currently it is checking for `>`.

PR #133, opened 2020-06-17
[Bug 1645897] Add descriptions to generated Glean schemas

scholtzan:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1645897

Descriptions now appear in generated schemas, see https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/57d236d038727fe573ccb3c8d5a945c8598df65b/schemas/org-mozilla-mozregression/usage/usage.1.bq#L340

### DIFF
--- a/mozilla_schema_generator/probes.py
+++ b/mozilla_schema_generator/probes.py
@@ -220,4 +220,8 @@ class GleanProbe(Probe):
     def get_schema(self, addtlProps: Any) -> Any:
         if addtlProps is None:
             raise SchemaException("Additional Properties cannot be missing for Glean probes")
+
+        if self.description is not None:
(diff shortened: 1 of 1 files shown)


mozilla_schema_generator/probes.py
```diff
@@ -220,4 +220,8 @@ def get_last_change(self) -> datetime:
     def get_schema(self, addtlProps: Any) -> Any:
         if addtlProps is None:
             raise SchemaException("Additional Properties cannot be missing for Glean probes")
+
+        if self.description is not None:
```
acmiyaguchi, line 224:
```suggestion
        if self.description:
```

PR #135, opened 2020-07-30
Add some description for the user_prefs field in the environment

wlach:
In particular, mentioning that only prefs in an allow list will
be collected.


wlach:
Unit tests need to be updated, apparently

PR #139, opened 2020-08-27
Inject mozPipelineMetadata to generated Glean json schemas

jklukas:
Makes Glean JSON schemas parallel to hand-written JSON schemas. See
https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/579

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,7 +38,7 @@ RUN pip install -r requirements/test_requirements.txt
 ADD . ${HOME}/mozilla-schema-generator
 ENV PATH $PATH:${HOME}/mozilla-schema-generator/bin
 
-RUN pip install -e ${HOME}/mozilla-schema-generator
+RUN pip install --no-dependencies -e ${HOME}/mozilla-schema-generator
 
 # Drop root and change ownership of the application folder to the user
 RUN chown -R ${USER_ID}:${GROUP_ID} ${HOME}
--- a/bin/alias_schemas
+++ b/bin/alias_schemas
@@ -57,7 +57,7 @@ def main(aliases_path, base_dir):
                 fname = source.name.replace(source_doctype, dest_doctype)
                 dest = dest_dir / fname
 
-                print(f"Aliasing {dest} to {source}")
+                print(f"Aliasing {source} to {dest}")
 
                 dest.parent.mkdir(parents=True, exist_ok=True)
                 shutil.copy(source, dest)
--- a/mozilla_schema_generator/__main__.py
+++ b/mozilla_schema_generator/__main__.py
@@ -214,16 +214,21 @@ def generate_glean_pings(config, out_dir, split, pretty, repo, generic_schema):
 
 
 def write_schema(repo, repo_id, config, out_dir, split, pretty, generic_schema):
-    schema_generator = GleanPing(repo)
+    schema_generator = GleanPing(repo, repo_id)
     schemas = schema_generator.generate_schema(config, split=False, generic_schema=generic_schema)
-    dump_schema(schemas, out_dir.joinpath(repo_id), pretty)
+    dump_schema(schemas, out_dir and out_dir.joinpath(repo_id), pretty)
 
 
 def dump_schema(schemas, out_dir, pretty, *, version=1):
-    json_dump_args = {'cls': SchemaEncoder}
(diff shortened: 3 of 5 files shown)


Dockerfile
```diff
@@ -38,7 +38,7 @@ RUN pip install -r requirements/test_requirements.txt
 ADD . ${HOME}/mozilla-schema-generator
 ENV PATH $PATH:${HOME}/mozilla-schema-generator/bin
 
-RUN pip install -e ${HOME}/mozilla-schema-generator
+RUN pip install --no-dependencies -e ${HOME}/mozilla-schema-generator
```
jklukas, line 41: This is an unrelated efficiency boost. We already install dependencies based on requirements.txt before this line, so we can save some time by avoiding validating dependencies here.

bin/alias_schemas
```diff
@@ -57,7 +57,7 @@ def main(aliases_path, base_dir):
                 fname = source.name.replace(source_doctype, dest_doctype)
                 dest = dest_dir / fname
 
-                print(f"Aliasing {dest} to {source}")
+                print(f"Aliasing {source} to {dest}")
```
jklukas, line 60: Unrelated bug fix.

mozilla_schema_generator/__main__.py
```diff
@@ -214,16 +214,21 @@ def generate_glean_pings(config, out_dir, split, pretty, repo, generic_schema):
...
+    json_dump_args = {
+        'cls': SchemaEncoder
+    }
     if pretty:
-        json_dump_args['indent'] = 4
-        json_dump_args['separators'] = (',', ':')
+        json_dump_args.update({
+            'indent': 4,
+            'separators': (',', ':'),
+            'sort_keys': True,
```
jklukas, line 230: Adding `sort_keys` here makes sure that `mozPipelineMetadata` appears near the top of the generated schemas, in line with how hand-written schemas from mps look.

mozilla_schema_generator/__main__.py
```diff
@@ -214,16 +214,21 @@ def generate_glean_pings(config, out_dir, split, pretty, repo, generic_schema):
 
 
 def write_schema(repo, repo_id, config, out_dir, split, pretty, generic_schema):
-    schema_generator = GleanPing(repo)
+    schema_generator = GleanPing(repo, repo_id)
```
jklukas, line 217: We need GleanPing to keep track of the repo_id (app_id) now since the generated metadata depends on it for the bq_dataset_family param.

jklukas:
Looks like I need to update tests

jklukas:
Tests are now passing, and you can see the effects of these logic changes in the following commit to `test-generated-schemas`:

https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/7a366cf2ef9dbe9eda6842ec40bf84c8badb0013

bin/alias_schemas
fbertsch: Hah, I remember writing this and being undecided about the order.

tests/test_glean.py
```diff
@@ -67,11 +67,17 @@ def test_generic_schema(self, glean, config):
 
         final_schemas = {k: schemas[k][0].schema for k in schemas}
         for name, schema in final_schemas.items():
+            meta = schema.pop("mozPipelineMetadata")
```
fbertsch, line 70: `print_and_test` makes missing/altered elements in JSON much easier to understand on failure. Can We instead explicitly update the `generic_schema` to include this section?

bin/alias_schemas
jklukas: I was definitely interpreting this as basically meaning "I've copied X source file to Y destination". But I can now see how you might think about aliasing the opposite way.

tests/test_glean.py
jklukas: I'll look into whether we can refactor this.

fbertsch: Just to be clear - I don't see a great way to update the class function. Instead just add this blob to `generic_schema` in this test, and compare it all using `print_and_test`.

bin/alias_schemas
fbertsch: Yeah, probably just needs a bit more explanatory text.

PR #140, opened 2020-08-27
Bug 1659750 Inject probes into account-ecosystem ping

jklukas:
Also see https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/599

See output in https://github.com/mozilla-services/mozilla-pipeline-schemas/compare/7a366cf2ef9dbe9eda6842ec40bf84c8badb0013...18dc0df62340383544cd81f209c74654b646b3d2

### DIFF
--- a/mozilla_schema_generator/common_ping.py
+++ b/mozilla_schema_generator/common_ping.py
@@ -47,32 +47,15 @@ class CommonPing(GenericPing):
             """Add a description to the types defined above."""
             return {**dtype, **dict(description=comment)}
 
-        active_addons = prepend_properties(("environment", "addons", "activeAddons")) \
-            + ("additionalProperties", "properties")
-
-        # If active_addons doesn't exist in the schema, then this is a simplified
-        # environment and we can skip the below modifications.
-        if schema.property_exists(active_addons):
-            schema.set_schema_elem(
-                prepend_properties(("environment", "settings", "userPrefs")),
-                with_description(
-                    string_map,
-                    "User preferences - limited to an allowlist defined in "
-                    "`toolkit/components/telemetry/app/TelemetryEnvironment.jsm`",
-                ),
-            )
-            schema.set_schema_elem(
-                    prepend_properties(("environment", "system", "os", "version")), string)
-            schema.set_schema_elem(
-                    prepend_properties(("environment", "system", "os", "hasSuperfetch")), boolean)
-            schema.set_schema_elem(
-                    prepend_properties(("environment", "system", "os", "hasPrefetch")), boolean)
-            schema.set_schema_elem(
-                    prepend_properties(("environment", "addons", "theme", "foreignInstall")),
-                    integer)
+        if schema.property_exists(prepend_properties(("environment", "addons"))):
+            active_addons = (prepend_properties(("environment", "addons", "activeAddons"))
+                             + ("additionalProperties", "properties"))
             schema.set_schema_elem(active_addons + ("foreignInstall",), integer)
             schema.set_schema_elem(active_addons + ("version",), string)
             schema.set_schema_elem(active_addons + ("userDisabled",), integer)
+            schema.set_schema_elem(
+                prepend_properties(("environment", "addons", "theme", "foreignInstall")),
+                integer)
             schema.set_schema_elem(
                 prepend_properties(("environment", "addons", "activeGMPlugins"))
                 + ("additionalProperties", "properties", "applyBackgroundUpdates"),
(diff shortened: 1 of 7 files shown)


mozilla_schema_generator/common_ping.py
```diff
@@ -49,33 +49,39 @@ def with_description(dtype: dict, comment: str) -> dict:
...
-        schema.set_schema_elem(active_addons + ("userDisabled",), integer)
-        schema.set_schema_elem(
-            prepend_properties(("environment", "addons", "activeGMPlugins"))
-            + ("additionalProperties", "properties", "applyBackgroundUpdates"),
-            with_description(
-                integer,
-                "Cast into an integer via mozilla-schema-generator. See bug 1611027.",
-            ),
-        )
+        if property_exists(schema.schema, active_addons):
```
jklukas, line 52: The `set_schema_elem` calls here are all the same, but we wrap groups of calls in logic that checks for the existence of the struct. This is necessary to support account-ecosystem, which has a limited environment compared to most common pings.

mozilla_schema_generator/utils.py
```diff
@@ -32,3 +32,15 @@ def prepend_properties(key: Tuple[str]) -> Tuple[str]:
     ```
     """
     return tuple(chain.from_iterable(zip(("properties" for k in key), key)))
+
+
+def property_exists(schema: Dict, key: Tuple[str]) -> bool:
```
fbertsch, line 37: This should live in `Schema`: `schema.property_exists(key: Tuple[str])`

mozilla_schema_generator/common_ping.py
```diff
@@ -50,32 +50,37 @@ def with_description(dtype: dict, comment: str) -> dict:
...
-            prepend_properties(("environment", "addons", "activeGMPlugins"))
-            + ("additionalProperties", "properties", "applyBackgroundUpdates"),
-            with_description(
-                integer,
-                "Cast into an integer via mozilla-schema-generator. See bug 1611027.",
-            ),
-        )
+        # If active_addons doesn't exist in the schema, then this is a simplified
+        # environment and we can skip the below modifications.
+        if property_exists(schema.schema, active_addons):
```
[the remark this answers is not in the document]
jklukas: Pipeline metadata is new enough that I'm not quite ready to think about how to annotate the environment field using that mechanism. But it would make sense to be a bit pickier here, and verify existence of specific sections of the environment before modifying them, rather than using active_addons as an indicator of a full environment.

Things get messy with the "os" modifications, though. I don't believe those are necessary and I'm pursing replacing them with fields in the source json schemas in https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/600

jklukas:
Tests are now passing again. Here is an exact diff of the generated content: https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/6a5dfdf7d3452fbcbbadb77ebb0ebf5f040e1f7b

mozilla_schema_generator/common_ping.py
```diff
@@ -41,41 +41,34 @@ def _update_env(self, schema):
...
-                string_map,
-                "User preferences - limited to an allowlist defined in `toolkit/components/telemetry/app/TelemetryEnvironment.jsm`",  # NOQA
-            ),
-        )
-        schema.set_schema_elem(
-                prepend_properties(("environment", "system", "os", "version")), string)
-        schema.set_schema_elem(
-                prepend_properties(("environment", "system", "os", "hasSuperfetch")), boolean)
-        schema.set_schema_elem(
-                prepend_properties(("environment", "system", "os", "hasPrefetch")), boolean)
```
jklukas, line 65: I have now removed these three environment.system.os changes. All 3 fields are now present in the "raw" schemas and I see no reason to inject them here anymore. This also allows the descriptions from the raw schemas to stay intact such that they are propagated to the BQ schemas.

mozilla_schema_generator/common_ping.py
```diff
@@ -41,41 +41,34 @@ def _update_env(self, schema):
...
+                prepend_properties(("environment", "addons", "activeGMPlugins"))
+                + ("additionalProperties", "properties", "applyBackgroundUpdates"),
+                with_description(
+                    integer,
+                    "Cast into an integer via mozilla-schema-generator. See bug 1611027.",
+                ),
+            )
+
+        user_prefs = (prepend_properties(("environment", "settings", "userPrefs")))
+        if schema.property_exists(user_prefs):
```
jklukas, line 68: We now check for environment.addons, and make all addon-related changes in that block. And then this block here checks for user_prefs and makes changes limited to that section.

This naturally exempts the limited AET environment from modifications.

PR #141, opened 2020-08-27
Use bq_metadata_format for choosing metadata structure

jklukas:
Depends on https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/598

This change should cause no difference to generated output, since the schemas now contain metadata to determine which format to use.

### DIFF
--- a/bin/metadata_merge
+++ b/bin/metadata_merge
@@ -2,13 +2,14 @@
 
 import click
 import json
+from os import path
 
 @click.command()
 @click.argument(
-    'metadata',
+    'metadata_dir',
     type=click.Path(
-        dir_okay=False,
-        file_okay=True,
+        dir_okay=True,
+        file_okay=False,
         writable=False,
         exists=True,
     ),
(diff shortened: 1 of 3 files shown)


acmiyaguchi:
Can you verify a clean diff on a test generated-schema branch with this change?

bin/metadata_merge
```diff
@@ -24,15 +25,27 @@ import json
     ),
     required=True
 )
-def main(metadata, schema):
-    print("Merging metadata {} and schema {}".format(metadata, schema))
-
-    with open(metadata, "r") as f:
-        metadata_contents = json.load(f)
+def main(metadata_dir, schema):
+    print("Merging metadata into schema {}".format(schema))
```
acmiyaguchi, line 29: nit: f-strings are nicer in most cases, imo

jklukas:
> Can you verify a clean diff on a test generated-schema branch with this change?

This is very nearly the right diff: https://github.com/mozilla-services/mozilla-pipeline-schemas/compare/7a366cf2ef9dbe9eda6842ec40bf84c8badb0013...b65369d100ff6d9dd62a39a45fa94790663404df

After https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/598 is merged, I'll work on a final validation here before merging.

jklukas:
And to be clear, the final diff will be no diff, because the generated content should be exactly the same. The machinery will decline to make a new commit because there are no changes. But again, I will triple-check that before merging.

jklukas:
"pioneer" is now present in master of m-p-s, and I have confirmed that the changes in this PR produce no change in generated output compared to current master of this repo.

whd (approved):

PR #142, opened 2020-08-28
Respect configured MPS_BRANCH_SOURCE for all URLs

jklukas:
Current behavior is that MPS_BRANCH_SOURCE is ignored for Glean pings and common pings, which always reference master.

I've confirmed that running `make build && make run` with these changes produces no difference in output, thus the code doesn't push a new commit to test-generated-schemas.

### DIFF
--- a/mozilla_schema_generator/generic_ping.py
+++ b/mozilla_schema_generator/generic_ping.py
@@ -24,13 +24,10 @@ class GenericPing(object):
     extra_schema_key = "extra"
     cache_dir = pathlib.Path(".probe_cache")
 
-    def __init__(self, schema_url, env_url, probes_url, mps_branch=None):
-        self.schema_url = schema_url
-        self.env_url = env_url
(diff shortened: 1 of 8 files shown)


mozilla_schema_generator/generic_ping.py
```diff
@@ -24,10 +24,13 @@ class GenericPing(object):
     extra_schema_key = "extra"
     cache_dir = pathlib.Path(".probe_cache")
 
-    def __init__(self, schema_url, env_url, probes_url):
+    def __init__(self, schema_url, env_url, probes_url, mps_branch=None):
         self.schema_url = schema_url
         self.env_url = env_url
         self.probes_url = probes_url
+        if mps_branch is not None:
```
fbertsch, line 31: This seems brittle (for when e.g. the url does not contain master), and also pushes the default branch now to the url rather than to the code here. Instead we should add a placeholder in the url, and always replace it with the branch, which can have a default value of `master`.

PR #146, opened 2020-09-08
Increase max column limit to 9600

fbertsch:


jklukas (approved):
LGTM modulo linting error.

PR #149, opened 2020-09-15
Remove first-shutdown from disallowlist

jklukas:
This appears to be an artifact of m-s-g before the concept of aliases was introduced. See https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/383

We have added saved-session as a main ping alias as well, but don't include it in the disallowlist. These should be symmetrical.


jklukas:
I won't merge this until schema deploys are back to being stable.

fbertsch (approved):

PR #150, opened 2020-09-16
Preserve mozPipelineMetadata in alias_schemas

jklukas:
~~Currently, all main ping clones get their metadata clobbered by main ping metadata.~~

This appears to be incorrect, but I'm not understanding yet how metadata is propagating.

### DIFF
--- a/bin/alias_schemas
+++ b/bin/alias_schemas
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
 import click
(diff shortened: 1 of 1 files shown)


jklukas:
I now understand how the metadata is preserved. I've filed https://github.com/mozilla/mozilla-schema-generator/pull/153 to better document what's going on.

PR #155, opened 2020-09-30
Copy bq schemas to a bq/ folder

jklukas:
See the ~[new generated content in the test-generated-schemas branch](https://github.com/mozilla-services/mozilla-pipeline-schemas/compare/db283cc5353d516922435d92a2ef12ecda524898...ea7ba2217223ef69248175e27309c285d5a88e48)~ (outdated)

### DIFF
--- a/bin/generate_commit
+++ b/bin/generate_commit
@@ -80,12 +80,11 @@ function _generate_schemas() {
     # Copy aliased BQ schemas into place
     alias_schemas "$ALIASES_PATH" .
 
-    popd
-
     # Copy all BQ schemas from schemas/ to appropriate paths under bq/
-    find schemas/ -name '*.bq' | \
+    find $mps_root/schemas/ -name '*.bq' | \
         xargs copy_to_bq_folder
 
+    popd
 }
 
 function _create_changelog() {
--- a/bin/generate_commit
+++ b/bin/generate_commit
@@ -79,7 +79,16 @@ function _generate_schemas() {
 
     # Copy aliased BQ schemas into place
     alias_schemas "$ALIASES_PATH" .
+
     popd
+
+    # Copy all BQ schemas from schemas/ to appropriate paths under bq/
+    # TODO: Remove the exemption for xfocsp-error-report/xfocsp-error-report
+    # once ops logic has adapted to the bq/ directory structure.
+    find schemas/ -name '*.bq' | \
+        grep -v xfocsp-error-report/xfocsp-error-report | \
+        xargs copy_to_bq_folder
+
 }
 
 function _create_changelog() {
@@ -127,7 +136,9 @@ function _commit_schemas() {
     # Note that in this step we throw out all the temporary modifications we've
     # made to the JSON schemas.
     find . -name "*.bq" -type f -exec git add {} +
-    git checkout ./*.schema.json
+
+    # Commit Glean JSON schemas and symlinks under bq/ dir
+    find . -name "*.schema.json" -exec git add {} +
 
     # Add Glean JSON schemas with generic schema
     mozilla-schema-generator generate-glean-pings \
@@ -142,9 +153,10 @@ function _commit_schemas() {
 
     git checkout "$mps_branch_publish" || git checkout -b "$mps_branch_publish"
 
-    # Keep only the schemas dir
+    # Keep only the schemas/ and bq/ dir
     find . -mindepth 1 -maxdepth 1 -not -name .git -exec rm -rf {} +
     git checkout "$mps_branch_working" -- schemas
+    git checkout "$mps_branch_working" -- bq
     git commit --all \
         --message "Auto-push from schema generation [ci skip]" \
         --message "$(_create_changelog "$mps_branch_source" "$mps_branch_publish")" \
--- a/bin/generate_commit
+++ b/bin/generate_commit
@@ -83,10 +83,7 @@ function _generate_schemas() {
     popd
 
     # Copy all BQ schemas from schemas/ to appropriate paths under bq/
-    # TODO: Remove the exemption for xfocsp-error-report/xfocsp-error-report
-    # once ops logic has adapted to the bq/ directory structure.
     find schemas/ -name '*.bq' | \
-        grep -v xfocsp-error-report/xfocsp-error-report | \
         xargs copy_to_bq_folder
 
 }
@@ -133,21 +130,22 @@ function _commit_schemas() {
     # This method will keep a changelog of releases. If we delete and newly
     # checkout branches everytime, that will contain a changelog of changes.
     # Assumes the current directory is the root of the repository.
-    # Note that in this step we throw out all the temporary modifications we've
-    # made to the JSON schemas.
     find . -name "*.bq" -type f -exec git add {} +
 
-    # Commit Glean JSON schemas and symlinks under bq/ dir
-    find . -name "*.schema.json" -exec git add {} +
+    # Throw out all the temporary modifications we've made to the JSON schemas.
+    git checkout ./*.schema.json
 
-    # Add Glean JSON schemas with generic schema
+    # Generate and stage Glean generic JSON schemas
     mozilla-schema-generator generate-glean-pings \
         --pretty \
         --generic-schema \
         --mps-branch "$mps_branch_source" \
         --out-dir schemas
+    find schemas/ -name "*.schema.json" -type f -exec git add {} +
 
-    find . -name "*.schema.json" -type f -exec git add {} +
+    # Stage BQ schemas and JSON schema symlinks under bq/ dir
+    find bq/ -name "*.schema.json" -exec git add {} +
+    find bq/ -name "*.bq" -exec git add {} +
 
     git commit -a -m "Interim Commit"
 
--- a/bin/copy_to_bq_folder
+++ b/bin/copy_to_bq_folder
@@ -3,11 +3,8 @@
 import click
 import json
 import os
-from pathlib import Path
+from pathlib import Path, PurePath
 import shutil
-
-ROOT_DIR = Path(__file__).parent.parent
-BQ_DIR = ROOT_DIR / "bq"
 
 
 @click.command()
@@ -29,13 +26,19 @@ def main(source_paths):
 
 def process_path(source_path):
     source_path = Path(source_path)
-    source_json_path = source_path.parent / (source_path.stem + '.schema.json')
-    source_bq_path = source_path.parent / (source_path.stem + '.bq')
+    stem = '.'.join(source_path.name.split('.')[:2])
+    source_json_path = source_path.parent / (stem + '.schema.json')
+    source_bq_path = source_path.parent / (stem + '.bq')
+    doctype_dir = source_path.parent
+    namespace_dir = doctype_dir.parent
+    schemas_dir = namespace_dir.parent
+    root_dir = schemas_dir.parent
+    bq_dir = root_dir / "bq"
 
     if source_json_path.exists() and source_bq_path.exists():
         print(f"Processing {source_bq_path} for copying to bq/")
     else:
-        print(f"Not copying {source_path} into {BQ_DIR} due to missing schema")
+        print(f"Not copying {source_path} into {bq_dir} due to missing schema")
         return
 
     with open(source_json_path, "r") as f:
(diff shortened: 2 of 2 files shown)


bin/schema_generator.sh
```diff
@@ -170,8 +173,16 @@ function main() {
     # Copy aliased BQ schemas into place
     alias_schemas $ALIASES_PATH .
 
+    cd ..
+
+    # Copy all BQ schemas from schemas/ to appropriate paths under bq/
+    # TODO: Remove the exemption for xfocsp-error-report/xfocsp-error-report
+    # once ops logic has adapated to the bq/ directory structure.
+    find schemas/ -name '*.bq' | \
+        grep -v xfocsp-error-report/xfocsp-error-report | \
```
whd, line 182: I think we're going to want to not exclude this. With this exclusion we have 0 bq schemas in the `bq` directory matching `xfocsp-error-report`. It really depends on the order in which we move/deprecate things, but in order for the ops pipeline to start using the `bq/` directory I'd expect the following script to have no output:
```sh
#!/bin/bash
cd schemas/
find -name '*.schema.json' -printf '%P\n' |
while read filename ; do
  namespace=$(echo $filename | cut -d/ -f1)
  if [[ $namespace =~ ^heka|metadata|glean|(pioneer-.*)$ ]] ; then
    continue
  fi
  f=$(cat $filename | jq -r .mozPipelineMetadata.bq_dataset_family)
  t=$(cat $filename | jq -r .mozPipelineMetadata.bq_table)
  if ! [[ -f ../bq/$f/${t}.bq ]]; then
    echo $filename
  fi
done
# output 
# xfocsp-error-report/xfocsp-error-report/xfocsp-error-report.4.schema.json
# telemetry/xfocsp-error-report/xfocsp-error-report.4.schema.json
```

Basically we want a 1:1 mapping of `.schema.json` to `.bq` files, and a 1:1 mapping of `.bq` files to tables. Though, even removing this exclusion, the PR will break the 1:1 mapping as well, as we have two xfocsp tables right now (one empty), with the plan to deprecate one of them.

Assuming the plan is to move `xfocsp-error-report` back into the `telemetry` `bq_dataset_family` once table ACLs are in place, some options here are:

1. Deprecate `xfocsp_error_report` fully before switching the ops logic to use the `bq` dir.
1. As  `bq_dataset_family` isn't used in gcp-ingestion or the ops logic yet, we could modify the telemetry xfocsp-error-report `bq_dataset_family` back to telemetry and do no exclusion here, relying on the rerouting special case in code until we've deprecated `xfocsp_error_report`. This decouples the ops logic adoption of `bq` from xfocsp dataset deprecation but leaves the metadata in a potentially confusing state.
1. Implement further special case logic in the ops pipeline when adopting the `bq` structure
We already have to have a [special case](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-shared/tf/modules/bigquery/generate_tfjson.sh#L291) for xfocsp rerouting due to the pipeline family mismatch and getting that into metadata is certainly out of scope for this phase. I think it would be reasonable to have extra special case logic during the transition to using `bq/`, but it would be ideal to avoid it.

I don't have a strong preference for any of the options above, and merging this PR as-is shouldn't break anything. I might have a more developed opinion when I'm actually implementing the ops changes.

whd (commented):
Modulo xfocsp this looks the way I'd expect; see the specific comment on that.

bin/generate_commit
```diff
@@ -79,7 +79,16 @@ function _generate_schemas() {
...
     alias_schemas "$ALIASES_PATH" .
+
     popd
+
+    # Copy all BQ schemas from schemas/ to appropriate paths under bq/
+    # TODO: Remove the exemption for xfocsp-error-report/xfocsp-error-report
+    # once ops logic has adapted to the bq/ directory structure.
+    find schemas/ -name '*.bq' | \
+        grep -v xfocsp-error-report/xfocsp-error-report | \
+        xargs copy_to_bq_folder
```
acmiyaguchi, line 90: You may want to do this before popping out of the schemas directory.

bin/schema_generator.sh
jklukas: I like option (2) best for now. I've filed https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/630 to remove the custom `bq_table` from `telemetry/xfocsp-error-report`. Once that is merged, I'll remove this exemption and rerun the logic to verify the output is as we expect.

bin/generate_commit
jklukas: This script is copying schemas from underneath `schemas/` into `bq/` which is a new top-level folder. It seems easier to reason about this operation from the parent directory under which both `schemas/` and `bq/` live rather than referencing `../bq` from within `schemas`.

acmiyaguchi: The function doesn't necessarily need to be run from the root of the docker application though. I would at least suggest referencing the `$mps_root` directory that's being passed in.

jklukas:
See new diff in https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/2c991e3db6cd3218332d4c843c0901e7333da5fb

whd:
> See new diff in [mozilla-services/mozilla-pipeline-schemas@2c991e3](https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/2c991e3db6cd3218332d4c843c0901e7333da5fb)

I'm seeing bad symlinks in `bq/` to json schemas with this new diff: https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/2c991e3db6cd3218332d4c843c0901e7333da5fb#diff-0b848acb770ceb68110bf3adf6865c99

jklukas:
I pushed a logic update that builds the symlink via `PurePath("../..") / source_json_path.relative_to(root_dir)`. See the [relevant diff for the generated results](https://github.com/mozilla-services/mozilla-pipeline-schemas/compare/a1cb05bd0c7831dd16fefcc328911ae3fd40c7f2...a6b33bb372cfc572574fed426f2b71f85ac6bc0b).

I pulled down the tarball and verified that the symlinks are now valid.

bin/generate_commit
jklukas: I've pulled this line above the `popd`.

acmiyaguchi (commented):
I noticed that there are some [files like this in the diff](https://github.com/mozilla-services/mozilla-pipeline-schemas/compare/a1cb05bd0c7831dd16fefcc328911ae3fd40c7f2...a6b33bb372cfc572574fed426f2b71f85ac6bc0b#diff-81402e5c1d3ed91252ef7ccdf01ff5c300f3ceaf60ec16d8d79250e61b06208cR1)

Is this just how github represents symlinks in the diff?

jklukas:
> Is this just how github represents symlinks in the diff?

Yes, github shows the text of the link for symlinks. I believe this mirrors what you get on a Windows machine when you clone a repository, since Windows doesn't support symlinks. If you download the tarball and untar it on a Mac, theses are untarred as symlinks.

acmiyaguchi (approved):
This looks to good to me. I see the `bq/` folder has been flattened by one level, which maps nicely into the way that BigQuery represents datasets and tables. +1

whd (approved):
Can you remind me what the eventual directory structure is we're expecting? I know there will be a set of intermediate/redundant representations as things are moved to the new structure. Is the final state what this PR produces with the `.bq` files under `schemas/` removed? Are these symlinks meant for humans to use or application code?

jklukas:
> Can you remind me what the eventual directory structure is we're expecting?

The `schemas/` directory will contain only `.schema.json` files; the `.bq` files will eventually be removed. To find the relevant `.bq` files, you'll be expected to parse `mozPipelineMetadata`, constructing the relevant path as `bq/<bq_dataset_family>/<bq_table>.bq`. I could, however, be convinced to instead convert the `.bq` files under `schemas/` to symlinks, which might make for more convenient human navigation.

> Are these symlinks meant for humans to use or application code?

I don't believe we currently have a machine use case for looking up the JSON schema corresponding to a given BQ schema, but if we did, my intention is that you would need to use the symlinks to determine which JSON schema corresponds to the given BQ table.

jklukas:
I'm holding off on merging this for the moment. It's gotten me doing more thinking about possibilities for the `bq/` structure. I've written up a [two-pager with an expanded vision for bq structure](https://docs.google.com/document/d/1ug7bP7-WJ6w0WypjDoSDWEdwav7suyJ6ECjGVIcRVgE/edit# ) and would very much like some first impression feedback on that.

jklukas:
I've updated this now quite a bit.

This now outputs files into directory paths that match the structure and filename conventions of bigquery-etl. For example, we now output a directory [`bq/moz-fx-data-shared-prod/activity_stream_stable/events_v1`](https://github.com/mozilla-services/mozilla-pipeline-schemas/tree/fefe3a773e9414cefc2c58bc3be966a08be42f92/bq/moz-fx-data-shared-prod/activity_stream_stable/events_v1). Note:

- That path contains project_id, dataset_id (including `_stable`), and table_id
- The contents include the BQ schema as `schema.yaml` (in YAML format) rather than as `event.1.bq` (in JSON format) since we have defined `schema.yaml` as the standard for capturing BQ schemas in bigquery-etl
- The contents now include a `metadata.yaml` with partitioning, clustering, expiration, etc.
- The `metadata.yaml` contains a new `source_schema` field that contains the relative path of the source JSON schema; there are no more symlinks used here

Some things still missing:
- This PR does not yet output tables under `_live` datasets, but we could do so easily if that's what we want here.
- This does not currently contain any ACL metadata; that probably should be held to a separate PR since we don't yet have an interface for defining ACLs in mozPipelineMetdata

@whd - Can you take a look at the format and let me know if this seems to be headed in a good direction? What's missing to make this bq/ folder serve as the relevant artifact for schema deploys?

whd:
I will look at this more thoroughly later today, but

> What's missing to make this bq/ folder serve as the relevant artifact for schema deploys?

Minimally, rally and pioneer schemas should be in a different project directory, as they target `moz-fx-data-pioneer-nonprod` and `moz-fx-data-pioneer-prod`.  As a further complication (not relevant to the current PR, which doesn't include user-facing datasets) from a technical perspective views datasets are published in pioneer/rally to specific analysis projects and not `moz-fx-data-pioneer-prod`, which is controlled by terraform state that this generation logic doesn't presently have access to.

I mentioned in email that it's difficult to support alternative target environments such as `moz-fx-data-shar-nonprod-efed` when hard-coding project names, but I'll think a bit more on how I think that should be handled.

whd (commented):
>     * That path contains project_id, dataset_id (including `_stable`), and table_id

This looks good, apart from aforementioned project id shenanigans.

>     * The contents include the BQ schema as `schema.yaml` (in YAML format) rather than as `event.1.bq` (in JSON format) since we have defined `schema.yaml` as the standard for capturing BQ schemas in bigquery-etl

This should be fine. I think this might completely abstract out `docversion` from ops logic, which in the end state would be ideal.

>     * The contents now include a `metadata.yaml` with partitioning, clustering, expiration, etc.

The question of labels arises. For ingestion datasets (`live`, `stable`) we don't typically specify them, but in particular `schemas_build_id` is something that can only really exist outside the commit. I think this may be a case where ops logic uses the artifact identifier (metadata about metadata heh) in deploy logic.
 
>     * The `metadata.yaml` contains a new `source_schema` field that contains the relative path of the source JSON schema; there are no more symlinks used here

This appears to be relative to the top-level `schemas` dir and not the root of the repo or the directory the file is in. That's probably fine, but I wanted to call it out. It implies that we'll be keeping that `schemas/` structure around; I'm not sure whether we decided we wanted to keep that directory structure around or if this is for compatibility while switching structures.

>     * This PR does not yet output tables under `_live` datasets, but we could do so easily if that's what we want here.

I think we want this.

>     * This does not currently contain any ACL metadata; that probably should be held to a separate PR since we don't yet have an interface for defining ACLs in mozPipelineMetdata

I agree. Some preliminary MPS ACL work was done in https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/646 but machinery for e.g. glean in this repo doesn't exist. Moving [tables_access](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-shared/tf/prod/envs/prod/bigquery/namespaces.tfvars.json#L558-L602) out of ops logic is a highly desirable end state however. It should be noted that without proper automation safety checks moving ACL management outside of cloudops-infra will require manual authorization of schemas deploys, so prioritizing that work will become important. I don't think that would block redundantly specifying the access in both places (e.g. having [here](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/test-generated-schemas/bq/moz-fx-data-shared-prod/telemetry_stable/deletion_v4/metadata.yaml) match [here](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-shared/tf/prod/envs/prod/bigquery/namespaces.tfvars.json#L585-L593) using the accepted metadata format from https://github.com/mozilla/bigquery-etl/pull/1628).

> @whd - Can you take a look at the format and let me know if this seems to be headed in a good direction? What's missing to make this bq/ folder serve as the relevant artifact for schema deploys?

We're obviously going to need to build on this to add dataset-level metadata, particularly for the glean all-channel `_derived` datasets that's motivating this work. I'd imagine we'd define all `_derived` datasets here too (along with live as previously mentioned), but looking at it now it might make sense to leave non-stable dataset definitions for later, since it will have some overlap with `namespaces.tfvars.json`, particularly datasets labeled `"ingestion_dataset": true`. It's probably also fine to redundantly specify in both systems and have ops logic pick and choose (or superset) configuration sources.

My plan here is to do the minimum amount of work on the old (tf 0.11) stack to support new datasets defined in this repo, while supporting the new format more robustly in the modern stack (currently only used to deploy pioneer/rally and mozdata).

PR #157, opened 2020-10-09
Add MPS_VALIDATE_BQ env var to allow disabling validation

jklukas:
When making changes to this repo, the testing workflow may involve pushing
changes to test-generated-schemas, and then needing to walk them back.

This PR adds a new environment variable that can be set to "false" to disable
the validation step and allow exploration with pushing to branches like
test-generated-schemas.

### DIFF
--- a/bin/schema_generator.sh
+++ b/bin/schema_generator.sh
@@ -12,6 +12,7 @@
 #   MPS_BRANCH_SOURCE:  The source branch for generating schemas e.g. master
 #   MPS_BRANCH_PUBLISH: The destination branch for publishing schemas
 #                       e.g. generated-schemas
+#   MPS_VALIDATE_BQ:    Set to 'false' to disable the validation step
 #
 # Example usage:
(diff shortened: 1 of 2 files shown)


bin/schema_generator.sh
```diff
@@ -92,9 +94,12 @@ function validated_generate_commit() {
 
     # Generate schemas, commits results to the publish branch
     generate_commit "$mps_root" "$mps_branch_source" "$mps_branch_publish"
-    validate-bigquery local-validation \
-        --head "$mps_branch_publish" \
-        --base "$mps_branch_temp"
+
+    if [ "$MPS_VALIDATE_BQ" != "false" ]; then
```
acmiyaguchi, line 98: It may also be worth considering `set +e` before and `set -e` after during the script based on the environment variable.

acmiyaguchi (approved):

bin/schema_generator.sh
jklukas: The implications of that are not completely clear to me, so I'm going to go ahead and merge this as-is (since it's relevant to concurrent work). But feel free to file a follow-up.

PR #158, opened 2020-10-14
Use glean-core as default dependency

fbertsch:
At this point, there are no repositories using the default dependency, so this shouldn't change anything.


mozilla_schema_generator/glean_ping.py
```diff
@@ -27,8 +27,7 @@ class GleanPing(GenericPing):
     repos_url = GenericPing.probe_info_base_url + "/glean/repositories"
     dependencies_url_template = GenericPing.probe_info_base_url + "/glean/{}/dependencies"
 
-    default_dependencies = ['glean']
-    ignore_pings = {"all-pings", "all_pings", "default", "glean_ping_info", "glean_client_info"}
```
fbertsch, line 31: I noticed that this was unused. Clearly we need to lint this repo.

jklukas:
Why was this closed? Are we still able to do this?

PR #161, opened 2020-10-15
Set default branch in commands to track mozilla-pipeline-schemas

### LINKED ISSUE #160: Add default option to generate-glean-pings command
When generating glean-ping-schemas for each app using the following command
```
mozilla-schema-generator generate-glean-pings
```

It throws this error
Screenshot 2020-10-15 at 3 50 04 AM

When this option ```--mps-branch master```  is specified it uses mozilla-pipeline-schemas' master branch  as a reference
to generate schemas from.  

Shouldn't it be the default option?

If it isn't need to be set the default option for some reason. I think we should update the README with this information.

acmiyaguchi:
This fixes #160 by setting the default branch to `master`. The `__main__` entrypoint is formatted and refactored so the default is set in a single place.

### DIFF
--- a/mozilla_schema_generator/__main__.py
+++ b/mozilla_schema_generator/__main__.py
@@ -23,42 +23,60 @@ CONFIGS_DIR = ROOT_DIR / "configs"
 SCHEMA_NAME_RE = re.compile(r".+/([a-zA-Z0-9_-]+)\.([0-9]+)\.schema\.json")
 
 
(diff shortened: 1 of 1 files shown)


fbertsch (approved):
So strange. I could have sworn this was working fine as-is earlier.

acmiyaguchi:
It's always called with an explicit branch in the main script, so the issue was probably hidden to us.

PR #163, opened 2020-10-17
Remove redundant conditional

DawoodShahat:


wlach (approved):
This makes sense, thanks!

PR #170, opened 2020-12-18
Bug 1682627 - Support propagating schema incompatible changes

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1682627)

This adds support for a incompatibility-allowlist that allows for deletions or schema incompatible changes. This is done by omitting the document from the base revision. The entries only need to exist in the allowlist for the single commit (and should probably be removed afterwards).

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,30 @@ This specific field is for indicating which table group that section of the ping
 splitting the schema. Currently we do not split any pings. See the section on [BigQuery
 Limitations and Splitting](#bigquery-limitations-and-splitting) for more info.
 
+## Allowing schema incompatible changes
+
+On every run of the schema generator, there is a check for incompatible changes
+between the previous revision and current generated revision. A schema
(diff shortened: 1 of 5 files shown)


jklukas (approved):
r+wc

This looks reasonable. The README development section should be updated to include a reference to this new allowlist.

I like that this approach is explicit about which particular doctypes should be allowed to be removed. It's a bit clunky that we'll end up with a history of outdated doctypes accumulated in that file over time. Aesthetically, I think I'd prefer a mechanism that relies on a special tag in the commit message (similar to CircleCI's `[ci-skip]` tag), but the solution here overall seems more easily explainable.

acmiyaguchi:
> It's a bit clunky that we'll end up with a history of outdated doctypes accumulated in that file over time.

We should probably clean up the file once the commit has been made to avoid leaving a hole in validation for too long. I've made note of this in both the README and the allowlist itself.

PR #173, opened 2021-01-05
Only include last N commits in generated-schemas README

### LINKED ISSUE #172: Embedded changelog causes noop deploys
I dislike the embedded changelog added in https://github.com/mozilla/mozilla-schema-generator/pull/171, as that changelog inconveniences machines.

I believe this change is going to introduce a secondary commit to `generated-schemas` ~28 days after each day an MPS commit occurs, even when there are no actual schemas changes. This will currently cause a redundant deploy on such days.

The standing assumption is that every commit to `generated-schemas` changes schemas (and should therefore propagate to GCP), which is easy for both humans and machines to reason about. With an embedded changelog in `generated-schemas` we either (a) redundantly redeploy bigquery resources even when schemas aren't changing, or (b) switch schemas deploy logic to only deploy changes to the branch that actually change schemas via e.g. hashing `git ls-tree`, making it more difficult to determine when schemas are up-to-date and when they are not.

The primary reason I would like to avoid these redundant bigquery deploys is because a 1:1 mapping between `generated-schemas` commits and schemas changes is very simple to reason about. I don't want to introduce commits for which the answer to "what generated schemas changed in this revision?" is "none". Secondarily, the BQ API returns legacy SQL and therefore terraform computes that every change requires updating the schema (see https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/bigquery_table#schema) making redeploys expensive (i.e. not just a label change). We rely on `schemas_build_id` to determine when to deploy schemas in most cases.

Aside: technically there are redundant shared-prod deploys now in cases where only pioneer/ion/rally schemas are updated (rally-prod is deployed manually), but this README has explicitly introduced a guaranteed category of commits that contain no schemas changes whatever. To be pedantic, these commits aren't guaranteed to be no-ops (e.g. [today's](https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/0236bbcfe9d0945a254721efa8973d010562fc83)) and in fact given our high schema update rate will perhaps only occasionally result in no-op commits, but it's still something I'd prefer not to introduce.

Can we just link to https://github.com/mozilla-services/mozilla-pipeline-schemas/commits/master or (`main` when it changes) instead? Alternatively, can we re-characterize the information in the README as "commits in the last 28 days from a schema change" and only update it when a schema is also being changed? As a matter of semantics, given the current MSG deploy cadence, "commits in the last 28 days" is going to be incorrect somewhat often for changes that are deployed on e.g. UTC Fridays.

fbertsch:
Fixes #172

Not tied to this approach, just putting this forward as an alternative.

### DIFF
--- a/bin/generate_commit
+++ b/bin/generate_commit
@@ -117,7 +117,7 @@ branch of [mozilla-pipeline-schemas](https://github.com/mozilla-services/mozilla
 See the [mps-deploys](https://protosaur.dev/mps-deploys/) dashboard for deployment status of schemas
 to [gcp-ingestion](https://github.com/mozilla/gcp-ingestion) and BigQuery.
 
-## Last 10 Changes
+## last 10 changes
 
 \`\`\`bash
 $(git log master --pretty=format:"%h%x09%cd%x09%s" --date=iso -n 10)
--- a/bin/generate_commit
+++ b/bin/generate_commit
@@ -109,8 +109,6 @@ function _create_changelog() {
 }
 
 function _create_readme() {
-    # date 28 days ago
-    start_date=$(python3 -c "import datetime as dt; print(str(dt.datetime.now()-dt.timedelta(28))[:10])")
     cat << EOF > README.md
 # mozilla-pipeline-schemas: generated-schemas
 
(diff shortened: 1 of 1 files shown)


acmiyaguchi (approved):
I'm also okay just removing the whole section for recent commits, but I believe this should solve the problem with noisy redeploys (which was wholly an unanticipated change).

bin/generate_commit
```diff
@@ -119,10 +117,10 @@ branch of [mozilla-pipeline-schemas](https://github.com/mozilla-services/mozilla
 See the [mps-deploys](https://protosaur.dev/mps-deploys/) dashboard for deployment status of schemas
 to [gcp-ingestion](https://github.com/mozilla/gcp-ingestion) and BigQuery.
 
-## commits in the last 28 days
+## Last 10 Changes
```
whd, line 120:
```suggestion
## last 10 changes
```
internal stylistic consistency

whd (approved):

PR #174, opened 2021-01-05
Remove commit section in README

acmiyaguchi:
I was thinking about https://github.com/mozilla/mozilla-schema-generator/issues/172 more. Even if we limit to the last N commits in the README in #173, there will still be no-op commits with changes to MPS that don't affect schemas (e.g. documentation or CI changes). If we want more detailed information, we should look to the commit log or the deploy dashboard for this.


fbertsch (approved):
No problem from me

PR #193, opened 2021-04-13
Optimize Dockerfile

jklukas:
The recursive `chown` step takes over 30 seconds to execute, significantly
slowing down the development process when attempting to test changes with
`make build && make run`.

By putting this step earlier, changes to the `bin/` content no longer trigger
a rerun of the `chown` step, cutting build time to less than 10 seconds.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,17 @@
+**/__pycache__/
+*.pyc
+*.pyo
+*.pyd
+.Python
+venv
+pip-log.txt
+pip-delete-this-directory.txt
+.tox
+.coverage
+.coverage.*
+.cache
+nosetests.xml
+coverage.xml
+*,cover
+*.log
+.git
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,17 +31,20 @@ RUN pip install --upgrade pip
 
 WORKDIR ${HOME}
 
-ADD requirements requirements/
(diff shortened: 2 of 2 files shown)


.dockerignore
```diff
@@ -0,0 +1,17 @@
+**/__pycache__/
```
jklukas, line 1: These temporary files were bloating the size of the context that's getting fed into image builds.

fbertsch (approved):
Good change, can you confirm that the schema generation process is successful?

jklukas:
> Good change, can you confirm that the schema generation process is successful?

I was able to successfully run `make build && make run` with these changes. It generated https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/c6fc1cccec74d631c786a3a1a9566176b9a24b5d (which includes a bunch of deletions under `bq/` due to builds I've been pushing for https://github.com/mozilla/mozilla-schema-generator/pull/155).

fbertsch:
Great, feel free to merge when you're ready.

jklukas:
Looks like [the build on `main` failed](https://app.circleci.com/pipelines/github/mozilla/mozilla-schema-generator/891/workflows/e6e020a1-fcc7-4b6a-bf0e-b4fbcb820fa3/jobs/1340), which must be a result of this PR. It's likely that my local builds were still using cache for the affected step. Will file a follow-up immediately.

PR #197, opened 2021-06-01
Add a method to get descriptions for an application's glean pings

wlach:
Needed for https://github.com/mozilla/lookml-generator/pull/143

### DIFF
--- a/mozilla_schema_generator/glean_ping.py
+++ b/mozilla_schema_generator/glean_ping.py
@@ -134,15 +134,22 @@ class GleanPing(GenericPing):
 
         return processed
 
-    def get_pings(self) -> Set[str]:
+    def _get_ping_data(self) -> Dict[str, Dict]:
         url = self.ping_url_template.format(self.repo_name)
-        pings = GleanPing._get_json(url).keys()
-
+        ping_data = GleanPing._get_json(url)
         for dependency in self.get_dependencies():
             dependency_pings = self._get_json(self.ping_url_template.format(dependency))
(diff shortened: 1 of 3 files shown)


mozilla_schema_generator/glean_ping.py
```diff
@@ -134,15 +134,22 @@ def get_probes(self) -> List[GleanProbe]:
...
             dependency_pings = self._get_json(self.ping_url_template.format(dependency))
-            pings |= dependency_pings.keys()
+            ping_data.update(dependency_pings)
+        return ping_data
+
+    def get_pings(self) -> Set[str]:
+        return self._get_ping_data().keys()
 
-        return pings
+    def get_ping_descriptions(self) -> Dict[str, str]:
```
fbertsch, line 148: Can we add a test for this in `tests/test_glean.py`?

wlach: Yep, I was just thinking of that after submitting this. Testing the full dependency resolution seems like a bit more than I really want to take on right now, but I added a quick test for the basic functionality (which actually caught a bug!)

wlach: Just to clarify the above comment a bit, lest I appear irresponsible: my preferred medium-term solution here is to provide better endpoints from probeinfo.tmo to reduce the complexity inside *this* repository.

fbertsch (approved):
I agree about the dependency resolution - I know we've discussed that before. For now though this should work.

PR #201, opened 2021-08-20
Use user's home directory to store probe cache

wlach:
The current working directory may not be readable by the user in all
cases (e.g. inside a docker container).

May help with: https://github.com/mozilla/bigquery-etl/issues/2214


wlach:
I'm not completely convinced this is the right solution, but it might be. Depends a bit on the deployment environment, so deferring to @whd for guidance (what directory is the failing script running from and does the user executing the script after write permissions to it?).

wlach:
Based on the traceback @whd just posted, this would solve mozilla/bigquery-etl#2214. However, I don't think we always want to dump this info in the user's home directory (e.g. for local development, we probably *do* want to dump the output in the current working directory). Instead, I'm going to change the behaviour to allow setting the probe cache dir via environment variable.

wlach:
Closing in favor of #202

PR #209, opened 2021-09-03
Process "in-source" annotations from probeinfo.tmo

wlach:
Will be useful in lookml-generator.


wlach:
While the PR that motivated this change (https://github.com/mozilla/lookml-generator/pull/233) probably needs a bit more work, I've done enough testing of this change by itself to be confident that this is ready to merge.

scholtzan (approved):

PR #212, opened 2021-11-04
Bug 1739239 - Error if there is a mismatch between BigQuery schema mappings and glean core schema

wlach:


mozilla_schema_generator/__main__.py
```diff
@@ -197,10 +197,28 @@ def generate_glean_pings(
     with open(config, "r") as f:
         config_data = yaml.safe_load(f)
 
-    config = Config("glean", config_data)
```
wlach, line 200: This variable was being rewritten, which was confusing

wlach:
I'm not sure if this is the best way of doing this, but it seems to work. Part of me feels guilty for not adding tests, but I'm not sure offhand to write them given the current architecture of MSG.

mozilla_schema_generator/__main__.py
```diff
@@ -197,10 +197,36 @@ def generate_glean_pings(
...
+    glean_metrics_in_schema = set(
+        glean_schema.get(["properties", "metrics", "properties"]).keys()
+    )
+    KNOWN_UNMATCHED_GLEAN_TYPES = {"url", "text", "labeled_rate", "jwe"}
+    new_unmatched_glean_types = (
+        glean_metrics_in_schema
+        - glean_matched_metrics_in_config
+        - KNOWN_UNMATCHED_GLEAN_TYPES
+    )
+    if len(new_unmatched_glean_types):
```
jklukas, line 213:
```suggestion
    if new_unmatched_glean_types:
```

I believe it's python idiom to rely on an empty collection itself being falsy rather than checking length.

mozilla_schema_generator/__main__.py
```diff
@@ -197,10 +197,36 @@ def generate_glean_pings(
...
+            "Unknown metric types in Glean Schema: {}. Please add them to {}".format(
+                ", ".join(new_unmatched_glean_types), config
+            )
+        )
 
     for repo in repos:
-        write_schema(repo, config, out_dir, split, pretty, generic_schema, mps_branch)
+        write_schema(
+            repo,
+            Config("glean", config_data),
```
relud, line 223: this is recreating the config for each repo, I think it would be better to rename it as `glean_config` or something but still instantiate it before this loop.

mozilla_schema_generator/__main__.py
```diff
@@ -197,10 +197,36 @@ def generate_glean_pings(
...
 
-    config = Config("glean", config_data)
+    # validate that the config has mappings for every single metric type specified in the
+    # Glean schema (see: https://bugzilla.mozilla.org/show_bug.cgi?id=1739239)
+    glean_schema = GleanPing(repos[0]).get_schema()
+    glean_matched_metrics_in_config = set(config_data["metrics"].keys())
+    glean_metrics_in_schema = set(
+        glean_schema.get(["properties", "metrics", "properties"]).keys()
+    )
+    KNOWN_UNMATCHED_GLEAN_TYPES = {"url", "text", "labeled_rate", "jwe"}
```
relud, line 207: I think this constant should be moved to line 25

jklukas:
This approach overall seems fine. And I agree that it's better to throw an error rather than silently ignoring new metric types that haven't made it into the config here.

relud:
> I'm not sure if this is the best way of doing this, but it seems to work. Part of me feels guilty for not adding tests, but I'm not sure offhand to write them given the current architecture of MSG.

I'll file a follow-up PR to add a test for this

jklukas (approved):

PR #213, opened 2021-11-05
Bug 1737656 Allow text and url types for new glean pings

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1737656


jklukas:
cc @wlach and @relud

The approach here is to keep a one-time list of which glean pings have already been deployed with the 4 incorrect types, and we remove the matchers for those existing types. I'm going to run the deploy-to-testing-branch workflow between main and this branch to validate there are zero changes in output. Then I'll remove one of the burnham doctypes from the list and validate that this can produce output with correct types.

mozilla_schema_generator/glean_ping.py
```diff
@@ -184,6 +184,23 @@ def generate_schema(
                     {"source_field_path": "/payload", "decrypted_field_path": ""}
                 ]
 
+            matchers = {
+                loc: m.clone(new_table_group=ping) for loc, m in config.matchers.items()
+            }
+
+            # See https://bugzilla.mozilla.org/show_bug.cgi?id=1737656
```
wlach, line 191: I'd like to see a longer description of what's going on here.

wlach (changes requested):
This seems like quite an elegant solution which should be enough to tide us over for the next while (or maybe indefinitely). We may need to coordinate with the Glean team to make it clear that these metric types are not currently supported for older applications / pings.

I know it's a pain, but I'd also like to see one or two tests of this logic to make sure this keeps working in the future.

jklukas:
Two test cases added.

mozilla_schema_generator/__main__.py
```diff
@@ -22,7 +22,6 @@
 ROOT_DIR = Path(__file__).parent
 CONFIGS_DIR = ROOT_DIR / "configs"
 SCHEMA_NAME_RE = re.compile(r".+/([a-zA-Z0-9_-]+)\.([0-9]+)\.schema\.json")
-KNOWN_UNMATCHED_GLEAN_TYPES = {"url", "text", "labeled_rate", "jwe"}
```
jklukas, line 25: This is no longer needed, since this PR adds these four types to the config.

wlach (approved):
Great stuff! Thank you

PR #214, opened 2021-11-08
Bug 1737656 Rename metrics.url to metrics.url2, etc.

jklukas:
This is one piece of the design discussed in https://bugzilla.mozilla.org/show_bug.cgi?id=1737656#c21

Relies on a coordinated pipeline change being developed in https://github.com/mozilla/gcp-ingestion/pull/1881

Although it should be possible to deploy these in any order without breaking the pipeline.

### DIFF
--- a/mozilla_schema_generator/configs/glean.yaml
+++ b/mozilla_schema_generator/configs/glean.yaml
@@ -95,24 +95,53 @@ metrics:
       send_in_pings:
         not:
           - glean_ping_info
-      type: labeled_rate
+      type: not_a_real_type
+
+  # Metric types affected by bug 1737656;
+  # The jwe, labeled_rate, url, and text types were deployed to BigQuery tables
+  # with incorrect structure, so we've had to rename these types in BQ tables;
+  # we use a "2" suffix for the fields in BQ.
   jwe:
     match:
       bug_1737656_affected: true
       send_in_pings:
         not:
           - glean_ping_info
-      type: jwe
+      type: not_a_real_type
+  labeled_rate2:
+    match:
+      send_in_pings:
+        not:
+          - glean_ping_info
+      type: labeled_rate
   text:
     match:
       bug_1737656_affected: true
       send_in_pings:
         not:
           - glean_ping_info
-      type: text
+      type: not_a_real_type
   url:
     match:
       bug_1737656_affected: true
+      send_in_pings:
+        not:
+          - glean_ping_info
+      type: not_a_real_type
+  jwe2:
+    match:
+      send_in_pings:
+        not:
+          - glean_ping_info
+      type: jwe
+  text2:
+    match:
+      send_in_pings:
+        not:
+          - glean_ping_info
+      type: text
+  url2:
+    match:
       send_in_pings:
         not:
           - glean_ping_info
--- a/mozilla_schema_generator/configs/glean.yaml
+++ b/mozilla_schema_generator/configs/glean.yaml
@@ -90,7 +90,7 @@ metrics:
           - glean_ping_info
       type: labeled_string
 
-  # Metric types affected by bug 1737656;
+  # Metric types affected by https://bugzilla.mozilla.org/show_bug.cgi?id=1737656
   # The jwe, labeled_rate, url, and text types were deployed to BigQuery tables
   # with incorrect structure, so we've had to rename these types in BQ tables;
   # we use a "2" suffix for the fields in BQ.
--- a/mozilla_schema_generator/configs/glean.yaml
+++ b/mozilla_schema_generator/configs/glean.yaml
@@ -89,19 +89,25 @@ metrics:
         not:
           - glean_ping_info
       type: labeled_string
-  labeled_rate:
-    match:
-      bug_1737656_affected: true
-      send_in_pings:
-        not:
-          - glean_ping_info
-      type: not_a_real_type
 
   # Metric types affected by bug 1737656;
   # The jwe, labeled_rate, url, and text types were deployed to BigQuery tables
   # with incorrect structure, so we've had to rename these types in BQ tables;
   # we use a "2" suffix for the fields in BQ.
   jwe:
+    match:
+      bug_1737656_affected: true
+      send_in_pings:
+        not:
+          - glean_ping_info
+      type: not_a_real_type
+  jwe2:
+    match:
+      send_in_pings:
+        not:
+          - glean_ping_info
+      type: jwe
+  labeled_rate:
     match:
       bug_1737656_affected: true
       send_in_pings:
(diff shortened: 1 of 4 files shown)


jklukas:
I _think_ this is functionally complete as-is, but I'd very much like to have a test in here that verifies we correctly emit fields like `metrics.url2.my_url` when a `url`-type metric is defined.

mozilla_schema_generator/configs/glean.yaml
```diff
@@ -95,24 +95,53 @@ metrics:
...
+  # we use a "2" suffix for the fields in BQ.
   jwe:
     match:
       bug_1737656_affected: true
       send_in_pings:
         not:
           - glean_ping_info
-      type: jwe
+      type: not_a_real_type
+  labeled_rate2:
```
jklukas, line 111: I've done enough testing to verify this is not sufficient to allow the rename. The code still expects the generic glean schema to contain these fields.

badboy: One thing I wonder: Why was `labeled_rate` affected here? That's not a thing we support anyway (afaik).

jklukas: Both `labeled_rate` and `jwe` were deployed to schemas but never supported in the SDK. I figure we might as well make the changes for all 4 affected types, even though we may never have a `labeled_rate` or `jwe` metric defined. As long as no apps define these metrics, `jwe2` and `labeled_rate2` will never actually appear in schemas.

badboy: Thanks.
And I take it that `rate` was unaffected because it was part of the initial list of metrics, without ever having been implemented (now I wonder if rates do properly work 🤔)

jklukas: Yes, the failure mode historically has been if a new metric is added to the glean JSON schema without being added to the `configs/glean.yaml` in this repository. AFAIK, `rate` was correctly in place here and so was never affected by the incorrect schema deploy issue.

jklukas:
This now contains a test that validates handling of a "url"-type metric being place in `metrics.url2`.

mozilla_schema_generator/glean_ping.py
```diff
@@ -57,6 +57,22 @@ def __init__(self, repo, **kwargs):  # TODO: Make env-url optional
             **kwargs,
         )
 
+    def get_schema(self) -> Schema:
+        schema = super().get_schema()
+
+        # We need to inject placeholders for the url2, text2, etc. types as part
+        # of mitigation for https://bugzilla.mozilla.org/show_bug.cgi?id=1737656
+        for metric_name in ["labeled_rate", "jwe", "url", "text"]:
```
jklukas, line 65: I'm going to move this somewhere else. By putting it here, this change will cause the deployed JSON schemas for glean apps used to validation to contain these fields, which we don't want. See https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/d07e70bfc445557c45f8016ddcbacb599557efda

jklukas: I've accomodated this by adding a `generic_schema` boolean parameter.

mozilla_schema_generator/glean_ping.py
```diff
@@ -57,6 +57,30 @@ def __init__(self, repo, **kwargs):  # TODO: Make env-url optional
...
+            metric1 = schema.get(
+                ("properties", "metrics", "properties", metric_name)
+            ).copy()
+            metric1 = schema.set_schema_elem(
+                ("properties", "metrics", "properties", metric_name + "2"),
+                metric1,
+            )
```
relud, line 80: I thought we were only going to add the 2 suffix for existing pings, but this definitely looks like it will do it for all pings?

wlach:
FYI I won't be able to review this until Monday, please don't block on me if it's important this land soon.

mozilla_schema_generator/glean_ping.py
jklukas: This adds in a placeholder to the glean json schema for all pings. When metrics get injected later on, these fields disappear if no metrics are defined.

But it is true that this change as written means that all pings that define a `url`-type metric will have that metric put in place under `url2`. Downstream logic will be more complex if it has to handle both the `url2` case and the `url` case with correct type. 

But I definitely see the argument that it's desirable for new pings and apps to have the fully "correct" structure. It's mostly just view generation that will be affected.

relud (approved):

mozilla_schema_generator/configs/glean.yaml
```diff
@@ -89,30 +89,59 @@ metrics:
         not:
           - glean_ping_info
       type: labeled_string
-  labeled_rate:
+
+  # Metric types affected by bug 1737656;
```
wlach, line 93: I would make this a link

```suggestion
  # Metric types affected by https://bugzilla.mozilla.org/show_bug.cgi?id=1737656
```

tests/test_glean.py
```diff
@@ -181,3 +203,22 @@ def test_bug_1737656_unaffected(self, config):
...
+        )
+        schemas = glean.generate_schema(config, split=False)
+
+        final_schemas = {k: schemas[k][0].schema for k in schemas}
+        schema = final_schemas.get("metrics")
+        assert "url" not in schema["properties"]["metrics"]["properties"].keys()
+        assert "url2" in schema["properties"]["metrics"]["properties"].keys()
+        assert list(
+            (schema["properties"]["metrics"]["properties"]["url2"]["properties"].keys())
+        ) == ["my_url"]
```
wlach, line 224: Seems like it might be simpler just to test for the exact representation here?

mozilla_schema_generator/glean_ping.py
wlach: Much as it pains me to say this, I'm leaning towards thinking it's probably better if we just use the `2` suffix for all the things. Realistically we're probably going to be stuck with this for a while and it's going to be easier to deal with if the behaviour is consistent.

However, this comment might do with some slight updating to make sure the current behaviour is clear:

https://github.com/mozilla/mozilla-schema-generator/blob/b21d2ad0374ccf58e7c071da946d8c9a76e0a9e5/mozilla_schema_generator/glean_ping.py#L191

I guess we're assuming that no one has actually landed any of these metric types since #213 was merged?

wlach (approved):
This makes sense to me. My main concern is ensuring that the decisions here are well documented, as it seems like the sort of thing that might be difficult to reason about over time.

mozilla_schema_generator/glean_ping.py
jklukas: We are assuming that no one has landed metrics of any of these 4 types over all time. And it's important that we get this merged before an app does try to implement any of these.

jklukas: I've updated the commentary in `glean_ping.py` to make the new "2" behavior clear.

jklukas:
I did a final confirmation with local testing. This change gives "Everything up-to-date" from the `git push` so no changes are made to current schemas.

PR #223, opened 2022-03-29
skipping rally_debug events validation for metrics_text field

kik-kik:
skipping rally_debug events validation for metrics_text field

This is because for this event type metrics["properties"]["text"] is not required.

### DIFF
--- a/tests/test_glean.py
+++ b/tests/test_glean.py
@@ -185,7 +185,14 @@ class TestGleanPing(object):
         final_schemas = {k: schemas[k][0].schema for k in schemas}
         for name, schema in final_schemas.items():
             # Only this static list of pings should have the incorrect schema for text
-            if name not in ["deletion-request", "demographics", "enrollment", "study-enrollment", "study-unenrollment", "uninstall-deletion"]:
+            if name not in [
+                "deletion-request",
+                "demographics",
+                "enrollment",
+                "study-enrollment",
+                "study-unenrollment",
+                "uninstall-deletion",
+            ]:
                 continue
 
             metrics_text = schema["properties"]["metrics"]["properties"]["text"]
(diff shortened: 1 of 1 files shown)


tests/test_glean.py
```diff
@@ -184,7 +184,12 @@ def test_bug_1737656_affected(self, config):
 
         final_schemas = {k: schemas[k][0].schema for k in schemas}
         for name, schema in final_schemas.items():
+            # text field for rally_debug events not expected
+            if schema["mozPipelineMetadata"]["bq_dataset_family"] == "rally_debug":
+                continue
```
jklukas, line 189:
```suggestion
            # text field for rally_debug events not expected
            if name not in ["deletion-request", "demographics", "enrollment", "study-enrollment", "study-unenrollment", "uninstall-deletion"]:
                continue
```

Looking through the context around this code one more time, I realize this is not the change needed.
The context for bug 1737656 is that we accidentally deployed an incorrect schema for `metrics.text` and we have a shim in this repo to maintain that incorrect schema for a static set of legacy pings.

All the rally-debug pings at the time were affected, but we've now added the new `events` ping which is _not_ in the static list and so should never have this incorrect `metrics.text` bit.

The check that appears here will be true for _all_ pings defined for rally-debug, so it effectively renders this test as a no-op.

Instead, I think we should encode the exact list of pings we expect from the static list. I expect that set to be the six types shown in https://github.com/mozilla-services/mozilla-pipeline-schemas/tree/generated-schemas/schemas/rally-debug besides `events`.

jklukas (changes requested):

tests/test_glean.py
jklukas: The config file that controls this is https://github.com/mozilla/mozilla-schema-generator/blob/main/mozilla_schema_generator/configs/bug_1737656_affected.txt#L229-L234

jklukas: I can't remember exactly what is returned for keys of the ping dict. It may end up being necessary to check `schema["mozPipelineMetadata"]["bq_table"]` instead.

jklukas (approved):

PR #228, opened 2022-06-13
Bug 1763467 enable the use of moz_pipeline_metadata 

gleonard-m:
Enable the use of moz_pipeline_metadata field from the probeinfo API

### DIFF
--- a/mozilla_schema_generator/glean_ping.py
+++ b/mozilla_schema_generator/glean_ping.py
@@ -176,6 +176,79 @@ class GleanPing(GenericPing):
     def get_pings(self) -> Set[str]:
         return self._get_ping_data().keys()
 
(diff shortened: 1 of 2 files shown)


jklukas (commented):
These changes are clear, well-factored, and well-documented.

It looks like CI is complaining about the `lint` step, so please take a look at that.

And we should make sure that there are no unexpected changes here when we do a full run. See testing instructions in https://github.com/mozilla/mozilla-schema-generator#development-and-testing for how to push to a testing branch. If we do that with the `main` content and then with these changes, we should get a nice single commit with all differences.

jklukas (approved):
I've confirmed that these changes have no effect on the generated output. This looks great, @gleonard-m!

PR #233, opened 2023-01-13
Add check to ensure all json schemas have matching bq schema

relud:
logic copied from https://github.com/mozilla/mozilla-schema-generator/blob/bb15ec09baf808f245b9a0ce0b6c68deb42ee327/bin/generate_commit#L65-L70

### DIFF
--- a/bin/generate_commit
+++ b/bin/generate_commit
@@ -108,6 +108,20 @@ function _create_changelog() {
         --date=iso
 }
 
+function _check_all_schemas_have_bq() {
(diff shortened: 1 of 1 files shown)


whd (approved):

whd:
It appears this ends up causing errors due to a larger set of schemas needing to be excluded. Maybe https://github.com/mozilla/mozilla-schema-generator/blob/main/disallowlist is the right list.

relud:
filed https://github.com/mozilla/mozilla-schema-generator/pull/234 to fix

PR #234, opened 2023-01-17
Handle disallowlist when checking bq schemas exist

relud:


bin/generate_commit
```diff
@@ -109,17 +109,18 @@ function _create_changelog() {
...
-        bq_out=${fname/schema.json/bq}
-        if ! test -e "$bq_out"; then
-          echo "Missing expected bq schema file: $bq_out" >&2
-          exit 1
-        fi
-    done
+    MISSING_BQ_SCHEMAS="$(
+      find . -type f -name "*.schema.json" |
+      grep -vf "$DISALLOWLIST" |
+      sed 's/schema.json/bq/' |
```
relud, line 115: just noticed that above on line 82 the disallowlist is applied to bq schemas:
```suggestion
      sed 's/schema.json/bq/' |
      grep -vf "$DISALLOWLIST" |
```

whd (approved):
I noticed a quoting thing but looks like it was fixed

PR #238, opened 2023-02-10
[Bug 1816136] Add mozillavpn to incompatibility list

scholtzan:
Related to https://github.com/mozilla/probe-scraper/pull/553

Glean.js doesn't have `baseline` and `metrics` tables, but since there was a bug mozillavpn wasn't considered as Glean.js app so these schemas got initially created. With the fix they get removed again.

### DIFF
--- a/incompatibility-allowlist
+++ b/incompatibility-allowlist
@@ -40,7 +40,8 @@ firefox-accounts.account-ecosystem.*
 # rally-attention-stream.*
 
 # Bug 1804256
-# Delete mozillavpn.metrics and mozillavpn.baseline after reverting changes to unify VPN apps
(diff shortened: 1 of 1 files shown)


incompatibility-allowlist
```diff
@@ -42,5 +42,5 @@ firefox-accounts.account-ecosystem.*
 # Bug 1804256
 # Delete mozillavpn.metrics and mozillavpn.baseline after reverting changes to unify VPN apps
 # This is a companion to https://github.com/mozilla/probe-scraper/pull/550
```
sean-rose, line 44: Probably worth updating this comment to mention the current situation as well (e.g. https://github.com/mozilla/probe-scraper/pull/553).

sean-rose (approved):
r+wc

PR #239, opened 2023-03-02
Bug 1815432 - allow crash ping deletion

relud:
cc @whd

### DIFF
--- a/incompatibility-allowlist
+++ b/incompatibility-allowlist
@@ -41,7 +41,21 @@ firefox-accounts.account-ecosystem.*
 
 # Bug 1804256
 # Delete mozillavpn.metrics and mozillavpn.baseline after fixing typo in https://github.com/mozilla/probe-scraper/pull/553
-# mozillavpn was treated as Glean.js app after typo got fixed, which resulted in these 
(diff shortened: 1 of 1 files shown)


incompatibility-allowlist
```diff
@@ -41,7 +41,21 @@ firefox-accounts.account-ecosystem.*
...
 # Delete mozillavpn.metrics and mozillavpn.baseline after fixing typo in https://github.com/mozilla/probe-scraper/pull/553
-# mozillavpn was treated as Glean.js app after typo got fixed, which resulted in these 
+# mozillavpn was treated as Glean.js app after typo got fixed, which resulted in these
 # ping schemas being removed (since they don't exist for Glean.js)
 mozillavpn.metrics.*
 mozillavpn.baseline.*
+
+# Bug 1815432
+# Delete unused glean crash pings for glean applications other than firefox-desktop
+firefox-desktop-background-update.crash.*
```
relud, line 51: this one isn't completely empty, but per discussion with chutten it should still be deleted

chutten (approved):
Does this make it hard to add crash pings to these products later, or just remove protections against accidentally deleting them?

relud:
this will be commented out after the deletion is completed, so that it won't interfere with future work

PR #240, opened 2023-03-03
Add Cache-Control: no-cache to google cloud cdn requests

relud:
note this won't actually work until https://mozilla-hub.atlassian.net/browse/DSRE-1215 is deployed, because unless it's in the list of cache bypass headers, cloud cdn will ignore cache-control headers in requests.


relud:
cc @chutten this will fix the issue where schemas were taking a day longer than expected to deploy

scholtzan (approved):

PR #244, opened 2023-06-07
Add support for creating pings as a subset of other pings

relud:


bin/alias_schemas
```diff
@@ -46,21 +46,32 @@ def main(aliases_path, base_dir):
...
+            source_doctype = alias_info["source-doctype"]
             source_dir = base_dir / source_namespace / source_doctype
             dest_dir = base_dir / dest_namespace / dest_doctype
 
             assert source_dir.exists()
 
-            for source in source_dir.glob('*.bq'):
-                fname = source.name.replace(source_doctype, dest_doctype)
-                dest = dest_dir / fname
+            for source in source_dir.glob(f"{source_doctype}.*.schema.json"):
```
relud, line 56: this was modified to alias the json schema, so that a non-generic json schema is available for first-shutdown and saved-session pings to have use counters split out of them.

akkomar (approved):

PR #247, opened 2023-06-20
Fix missing escape and require --out-dir to generate subset pings

relud:


mozilla_schema_generator/__main__.py
```diff
@@ -252,8 +252,11 @@ def generate_subset_pings(config, out_dir, split, pretty, mps_branch):
     """
     if split:
         raise NotImplementedError("Splitting of subset pings is not supported.")
-    if out_dir:
-        out_dir = Path(out_dir)
+    if not out_dir:
+        raise NotImplementedError(
+            "Generating subset pings without out dir is not supported."
```
scholtzan, line 257:
```suggestion
            "Generating subset pings without out_dir is not supported."
```

scholtzan (approved):

PR #253, opened 2023-09-13
Remove split argument from generate_subset_pings

relud:
fixes bug introduced in #245


relud:
merging without review to fix overnight job

PR #254, opened 2023-09-13
Fix return type for subset_pings.generate

relud:
fixes bug introduced in https://github.com/mozilla/mozilla-schema-generator/pull/245


relud:
merging without review to fix overnight job

PR #258, opened 2023-12-12
add mozillavpn-cirrus to schema generator's incompatibility allowlist

Marlene-M-Hirose:
Part 1 of changing mozillavpn_cirrus to mozillavpn_backend_cirrus

### DIFF
--- a/incompatibility-allowlist
+++ b/incompatibility-allowlist
@@ -65,6 +65,6 @@ telemetry.main-remainder.*
 telemetry.saved-session-remainder.*
 telemetry.first-shutdown-remainder.*
 
- # Bug 1867770
- # Delete mozillavpn_cirrus so we can rename mozillavpn_cirrus to _mozillavpn_backend_cirrus
- mozillavpn_cirrus*
+# Bug 1867770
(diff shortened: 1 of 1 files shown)


incompatibility-allowlist
```diff
@@ -64,3 +64,7 @@ mozillavpn.baseline.*
 telemetry.main-remainder.*
 telemetry.saved-session-remainder.*
 telemetry.first-shutdown-remainder.*
+
+ # Bug 1867770
+ # Delete mozillavpn_cirrus so we can rename mozillavpn_cirrus to _mozillavpn_backend_cirrus
+ mozillavpn_cirrus*
```
sean-rose, line 70: - No leading spaces.
- The pipeline schemas use dashes for word separators, as you can see [here](73a4a7aea077f9cc3c6a1cfe6949b90c610eb5c5).
- `.*` instead of `*`.
```suggestion
# Bug 1867770
# Delete mozillavpn-cirrus so we can rename the mozillavpn_cirrus app to mozillavpn_backend_cirrus
mozillavpn-cirrus.*
```

sean-rose (approved):
r+wc

PR #260, opened 2024-01-05
[Bug 1870317] Handle changing probe types

scholtzan:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1870317

We have had a case where the type of an existing metric got changed in the metrics.yaml.
In this specific case a string metric web.page_url got changed to an url metric (see [PR](https://github.com/MozillaSocial/elk/pull/94/files)).

This PR uses the probe history to create a field for every type. The generated schemas contain an entry for every type that was seen in the `history`. 

The generated test schemas for mastodon have `web.page_url` as both `url` and `string`: https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/c44b9c8b982718a518576d8220db63a66f480ede/schemas/moso-mastodon-web/events/events.1.bq#L407-L435

### DIFF
--- a/tests/test_glean.py
+++ b/tests/test_glean.py
@@ -215,6 +215,93 @@ class TestGleanPing(object):
             }
             assert glean.get_ping_descriptions() == {"foo": "baz"}
 
+    def test_dependencies(self, glean):
+        RETURN_VALUES = {"glean-core": {"name": "glean-core", "type": "dependency"}}
+        with patch.object(
+            glean,
+            "_get_json",
+            return_value=RETURN_VALUES,
+        ):
+            assert set(glean.get_dependencies()) == {
+                "glean-core",
+            }
+
+    def test_probes(self, glean):
+        RETURN_VALUES = {
+            "test.probe": {
+                "history": [
+                    {
+                        "dates": {
+                            "first": "2023-11-27 21:59:53",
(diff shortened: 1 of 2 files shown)


badboy (approved):

tests/test_glean.py
```diff
@@ -215,6 +215,93 @@ def test_pings(self, glean):
...
+                        },
+                        "description": "description",
+                        "expires": "never",
+                        "type": "url",
+                        "version": 0,
+                    },
+                ],
+                "in-source": True,
+                "name": "test.probe",
+                "type": "string",
```
badboy, line 288: Ok, great. I typed up a whole response here and then GitHub just didn't save that or something.

So here it is again:
Initially I thought this should be `"type": "url"`, assuming that probeinfo service would do the right thing already.
But it does not: https://probeinfo.telemetry.mozilla.org/glean/moso-mastodon-web/metrics
Look for `page_url`.

The dictionary does list that metrics as URL though, so I guess it picks the last seen value:
https://dictionary.telemetry.mozilla.org/apps/moso_mastodon_web/metrics/web_page_url

Changing probeinfo then is followup work I guess

scholtzan: I created an issue for this here: https://github.com/mozilla/probe-scraper/issues/677

PR #261, opened 2024-03-12
[DENG-3022] Add support for minimum glean ping schema selection

quiiver:
By using the `include_info_section` in the latest ping history we switch between `glean` and `glean-min` schemas which are created by https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/808


tests/test_glean.py
sean-rose, line 1: Suggest adding a test to cover the `include_info_sections = False` case as well.

badboy (approved):

scholtzan (approved):

PR #263, opened 2024-04-25
Don't fall back to glean-core if no dependencies are found for a repository

akkomar:
Requires https://github.com/mozilla/mozilla-schema-generator/pull/262

### DIFF
--- a/mozilla_schema_generator/glean_ping.py
+++ b/mozilla_schema_generator/glean_ping.py
@@ -39,7 +39,7 @@ class GleanPing(GenericPing):
         GenericPing.probe_info_base_url + "/glean/{}/dependencies"
     )
 
-    default_dependencies = ["glean-core"]
+    default_dependencies = []
 
     with open(BUG_1737656_TXT, "r") as f:
         bug_1737656_affected_tables = [
@@ -92,7 +92,7 @@ class GleanPing(GenericPing):
                 self.dependencies_url_template.format(self.repo_name)
             )
         except HTTPError:
-            logging.info(f"For {self.repo_name}, using default Glean dependencies")
+            logging.info(f"No dependencies found for {self.repo_name}")
             return self.default_dependencies
 
         dependency_library_names = list(dependencies.keys())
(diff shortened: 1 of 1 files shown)


mozilla_schema_generator/glean_ping.py
```diff
@@ -92,7 +92,7 @@ def get_dependencies(self):
                 self.dependencies_url_template.format(self.repo_name)
             )
         except HTTPError:
-            logging.info(f"For {self.repo_name}, using default Glean dependencies")
+            logging.info(f"No dependencies found for {self.repo_name}")
```
badboy, line 95: A http error here means something went entirely wrong. Should we re-raise the exception here instead of silently failing?

mozilla_schema_generator/glean_ping.py
```diff
@@ -109,7 +109,7 @@ def get_dependencies(self):
                 dependencies.append(repos_by_dependency_name[name])
 
         if len(dependencies) == 0:
-            logging.info(f"For {self.repo_name}, using default Glean dependencies")
+            logging.info(f"No dependencies found for {self.repo_name}")
             return self.default_dependencies
```
badboy, line 113: Given `self.default_dependencies` is empty anyway now, this whole check can go. We return the empty `dependencies` a few lines further down anyway.

badboy (approved):
I tried tracing back why we added defaults there to begin with.
It looks like it's always been there: https://github.com/mozilla/mozilla-schema-generator/commit/9d27dcc1#diff-5ea6e4f5d0d4e8f3fdf30f77a9c31c0ac9bb2047fe9b32d690c319e91cf50342R20

I still don't know why we went with that instead of explicitly specfying glean as a dependency, but I could imagine that we just couldn't come up with "no deps, but also not proper Glean"-use case back then.

Thanks for checking the logs. Given these days we're very explicit about dependencies and empty dependencies means empty dependencies I'm good with this change.

mozilla_schema_generator/glean_ping.py
akkomar: I'm not sure why we're catching this exception here, we don't do this in any other place `_get_json` is called. 
This method returns an empty dictionary for repositories that do not have dependencies defined. Maybe in the past probe-info-service returned an error in such cases?

I checked logs further back in time - I don't see this message logged in the last 365 days (except for `accounts-backend` and `moso-mastodon-backend`). Because the same message is logged in two places I cannot say where was it logged, but my testing suggests these were all from couple of lines below.

I think it makes sense to not catch this exception here and let it bubble up to the calling method. I'll push a commit with this change.

mozilla_schema_generator/glean_ping.py
akkomar: 👍

PR #272, opened 2024-06-27
#1898105 Only use latest value for include_info_section

quiiver:
Rolling back the change to check the full history for `include_info_sections`. The new plan is to no longer allow the migrations from `include_info_sections: true` -> `include_info_sections: false` 

Also added the offending ping to the incompatibility allowlist


whd:
The plan is to roll this out tomorrow morning Pacific.

PR #266, opened 2024-05-29
Set glean app-level metadata for dependency pings

BenWu:
fixes https://github.com/mozilla/mozilla-schema-generator/issues/264

This applies `moz_pipeline_metadata` defined for the app to pings defined in dependencies. e.g. firefox_desktop.pageload

~~Merging this is dependent on a resolution for https://mozilla-hub.atlassian.net/browse/DSRE-1639 so we don't unexpectedly delete data for pageload~~

Ready for review now that https://github.com/mozilla/probe-scraper/pull/777 has been merged.

I tested this with `mozilla-schema-generator generate-glean-pings --out-dir out_dir` and the before and after diff is here https://github.com/mozilla/mozilla-schema-generator/commit/b5a68d088eb7a716fa3c5de94b54fe91d41818ff
The only difference is the `{"delete_after_days": 10000}` in pageload which seems correct

PR #280, opened 2024-08-21
Prevent glean metadata overrides form overwriting default metadata

BenWu:
https://github.com/mozilla/mozilla-schema-generator/pull/266 introduced a bug where the metadata for a ping set at the app-level might overwrite the app defaults in `moz_pipeline_metadata_defaults`.  This is because `GleanPing.apply_default_metadata` is called twice and the first call puts the `default_metadata` dict in the ping and then the second call will update that dict in place so the `default_metadata` will then have new values.

For example, this can be seen in [`pine.deletion-request`](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/4be954bed819451f970c026a44b5c07c68ce18ba/schemas/pine/deletion-request/deletion-request.1.schema.json#L11) where the retention is set to 1130 but it's set to 10000 in [probe-scraper](https://github.com/mozilla/probe-scraper/blob/main/repositories.yaml#L525).

I made a pr against the generated schemas to show the changes this would make on the schemas https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/827.  I'll go through those and make sure those are correct and won't unexpectedly delete things.  All the changes are decreases in retention period except for pine deletion-requests and pageload which means that this bug shouldn't have deleted anything unexpectedly.

### DIFF
--- a/mozilla_schema_generator/glean_ping.py
+++ b/mozilla_schema_generator/glean_ping.py
@@ -4,6 +4,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import copy
 import logging
 from pathlib import Path
 from typing import Dict, List, Set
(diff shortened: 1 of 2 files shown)


akkomar (commented):
LGTM, leaving final r+ to whd since we don't yet know if this won't delete anything (or re-request me pls).

BenWu:
I looked through https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/827 and the new settings look to match probe-scraper.  The only issue is that the firefox ios sync pings were added exactly 400 days ago and the default retention for the app is 400 so partitions would start getting dropped from this change.  We should manually set the retention for those in probe scraper before making this fix.  I commented on each of the affected pings in the other pr

whd (approved):

whd:
https://github.com/mozilla/probe-scraper/pull/795/files will land today after which I think it's safe to propagate this.

BenWu:
I'll merge this and rerun schema generator and make sure it's working before the end of the day

PR #286, opened 2025-02-18
fix(Bug #1944865): provide a configuration to force glean min ping on a ping by ping basis

quiiver:
not clear on the downstream repercussions


quiiver:
Closing this after a discussion on slack.

PR #314, opened 2026-03-12
[DENG-10780] Update GHA workflow

BenWu:
https://mozilla-hub.atlassian.net/browse/DENG-10780

pin actions to hashes, use env vars instead of templates in commands, put secret in environment, only run ci for open PRs so it doesn't run twice per push to branches with open PRs


sean-rose (approved):
r+wc

PR #315, opened 2026-04-01
[DENG-10877] Add fenix_nightly and fennec_aurora to glean_v2_to_v1 allowlist

BenWu:
These tables have been validated ([DENG-10876](https://mozilla-hub.atlassian.net/browse/DENG-10876)) so they are ready to be swapped over.  Full plan in https://mozilla-hub.atlassian.net/browse/DENG-10877

This is on top of the changes from https://github.com/mozilla/mozilla-schema-generator/pull/321 so the app just need to be removed from the allowlist and added to the incompatibility allowlist

### DIFF
--- a/incompatibility-allowlist
+++ b/incompatibility-allowlist
@@ -123,6 +123,20 @@ telemetry.first-shutdown.4
 telemetry.saved-session.*
 telemetry.saved-session-use-counter.*
 
+# DENG-10876
+org-mozilla-fenix-nightly.metrics.1
+org-mozilla-fenix-nightly.sync.1
+org-mozilla-fenix-nightly.first_session.1
+org-mozilla-fenix-nightly.adjust_attribution.1
+org-mozilla-fenix-nightly.health.1
+org-mozilla-fenix-nightly.captcha_detection.1
+org-mozilla-fennec-aurora.metrics.1
+org-mozilla-fennec-aurora.sync.1
+org-mozilla-fennec-aurora.captcha_detection.1
+org-mozilla-fennec-aurora.first_session.1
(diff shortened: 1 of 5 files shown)


akkomar (changes requested):
> `incompatibility-allowlist` doesn't need to be updated because tables will be manually deleted while deploys are paused

BQ schema validation diffs schemas against current `generated-schemas` branch, so these namespaces need to be added to incompatibility allowlist for the plan in https://github.com/mozilla/bigquery-backfill/pull/61/changes#diff-9e2d2724a25fe1e132eff1c469f817b03d23b3873b99d92fa6a708b0307fa394R81 to work.

BenWu:
> BQ schema validation diffs schemas against current `generated-schemas` branch

When this gets merged and then rerun, the generated-schemas branch will get updated. Then the tables will be deleted/updated per the plan and the jenkins deploys happen after. Wouldn't that make the schemas match?

akkomar:
> When this gets merged and then rerun, the generated-schemas branch will get updated.

Sorry, I wasn't clear, there's validation against that branch before its pushed which I had in mind: https://github.com/mozilla/mozilla-schema-generator/blob/main/bin/schema_generator.sh#L99

BenWu:
Rebased on to https://github.com/mozilla/mozilla-schema-generator/pull/321 which changed how the v2->v1 gating worked and added to the incompatibility allowlist.

The schema diff in https://github.com/mozilla/mozilla-schema-generator/commit/b94ed40be16f017d7f6ab82f0ed32df5ec250ebf looks correct to me

akkomar (approved):

PR #317, opened 2026-05-14
[DENG-11139] Add CI workflow to generate schema diff

### LINKED ISSUE #216: Show a diff of any changes to generated schemas
To prevent problems like https://bugzilla.mozilla.org/show_bug.cgi?id=1737656 from recurring, we should use a similar approach as mozilla-pipeline-schemas, and show what any changes to the schema as a result of a push might look like via a ci bot.

The rough procedure we want to automate is here:

https://github.com/mozilla/mozilla-schema-generator#development-and-testing

Off the top of my head, I can think of two alternatives for getting the schemas to compare against (perhaps @jklukas can think of others):

1. Compare against running the schema generation process against main (basically what the instructions say to do right now).
2. Compare against what's in mozilla-pipeline-schemas generated-schemas branch.

## Prior Art

Example of MPS's CI bot in action:

https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/646#issuecomment-741181059

The implementation, I believe, is defined here:

* https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/main/.circleci/post-artifact.js
* https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/cd026852f61f119d721f2029041e2fd3caacd667/.circleci/config.yml#L49

BenWu:
https://mozilla-hub.atlassian.net/browse/DENG-11139

Also noticed that this closes #216

Generates schemas using the PR branch and main if there are potential changes and then commits them to a new branch so the diff can be viewed since doing this manually is tedious. [example](https://github.com/mozilla/mozilla-schema-generator/commit/df759febb752ab87d5d4e2cafb6e1274271cdeab)

I decided to rely on the `added a commit that referenced this pull request` notification instead of posting a comment each time to reduce noise.


tests/test_validate_bigquery.py
```diff
@@ -83,9 +83,7 @@ def test_copy_schemas(tmp_path, tmp_git):
     assert len(bq) > 0, "no bq schemas detected"
     assert len(bq) == len(txt)
     repo = Repo(tmp_git)
-    # the dst name encodes the revision, we should always be able to use the
-    # first 6 characters to find commit in the repository.
-    assert repo.rev_parse(dst.name[:6]) == repo.head.commit
+    assert repo.rev_parse(dst.name) == repo.head.commit
```
BenWu, line 86: There is a hash collision with the current latest generated-schemas commit so switching this to the full hash.  It doesn't need to be readable anyway

in m-p-s `git cat-file -t b5327b`:
```
error: short object ID b5327b is ambiguous
hint: The candidates are:
hint:   b5327b512 commit 2026-05-15 - Auto-push from schema generation [ci skip]
hint:   b5327b07a tree
fatal: Not a valid object name b5327b
```

.github/workflows/schema-diff.yml
```diff
@@ -0,0 +1,163 @@
...
+
+          DIFF_COMMIT_SHA=$(git rev-parse HEAD)
+          echo "diff_commit_sha=$DIFF_COMMIT_SHA" >> "$GITHUB_OUTPUT"
+          {
+            echo "## Generated Schema Diff Branch"
+            echo ""
+            echo "Pushed to [\`$DIFF_BRANCH\`](https://github.com/${GITHUB_REPOSITORY}/tree/${DIFF_BRANCH})."
+            echo ""
+            echo "View this PR's schema changes: [commit ${DIFF_COMMIT_SHA:0:7}](https://github.com/${GITHUB_REPOSITORY}/commit/${DIFF_COMMIT_SHA})"
+          } >> "$GITHUB_STEP_SUMMARY"
```
scholtzan, line 163: If there are multiple PRs running can they overwrite each others changes on the branch?

BenWu: They get overwritten since the branch only ever contains the latest two commits, but it still works because it's a single push with both commits.  The commits don't need to be on the branch to view the diff, it will just show `This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository.`

scholtzan (approved):

PR #327, opened 2026-08-12
[DENG-8494] Add desktop and fenix to v2 allowlist, update metrics blocklist

BenWu:
Creating v2 tables for desktop and fenix. Not doing much yet, but I want to have the schemas to test https://mozilla-hub.atlassian.net/browse/DENG-11473 and redo performance test.

List of table is from 
```sql
WITH glean_ping_2_tables AS (
  SELECT 'firefox_desktop_stable' AS table_schema, table_name
  FROM `moz-fx-data-shared-prod.firefox_desktop_stable.INFORMATION_SCHEMA.TABLE_OPTIONS`
  WHERE option_name = 'labels' AND option_value LIKE '%STRUCT("schema_id", "glean_ping_2")%'
  UNION ALL
  SELECT 'org_mozilla_firefox_stable', table_name
  FROM `moz-fx-data-shared-prod.org_mozilla_firefox_stable.INFORMATION_SCHEMA.TABLE_OPTIONS`
  WHERE option_name = 'labels' AND option_value LIKE '%STRUCT("schema_id", "glean_ping_2")%'
  UNION ALL
  SELECT 'org_mozilla_fenix_stable', table_name
  FROM `moz-fx-data-shared-prod.org_mozilla_fenix_stable.INFORMATION_SCHEMA.TABLE_OPTIONS`
  WHERE option_name = 'labels' AND option_value LIKE '%STRUCT("schema_id", "glean_ping_2")%'
  UNION ALL
  SELECT 'org_mozilla_firefox_beta_stable', table_name
  FROM `moz-fx-data-shared-prod.org_mozilla_firefox_beta_stable.INFORMATION_SCHEMA.TABLE_OPTIONS`
  WHERE option_name = 'labels' AND option_value LIKE '%STRUCT("schema_id", "glean_ping_2")%'
),
distribution_tables AS (
  SELECT DISTINCT table_schema, table_name
  FROM `moz-fx-data-shared-prod.region-us.INFORMATION_SCHEMA.COLUMN_FIELD_PATHS`
  WHERE table_schema IN ('firefox_desktop_stable', 'org_mozilla_firefox_stable', 'org_mozilla_fenix_stable', 'org_mozilla_firefox_beta_stable')
    AND field_path IN (
      'metrics.timing_distribution',
      'metrics.labeled_timing_distribution',
      'metrics.memory_distribution',
      'metrics.labeled_memory_distribution',
      'metrics.custom_distribution',
      'metrics.labeled_custom_distribution'
    )
)
SELECT
  d.table_schema || '.' || d.table_name AS table_name,
FROM distribution_tables d
LEFT JOIN glean_ping_2_tables g
  USING (table_schema, table_name)
WHERE g.table_name IS NULL
```

And metrics blocklist is updated to get everything that was removed >1 year ago as of 2026-08-01

### DIFF
--- a/bin/generate_metric_blocklist.py
+++ b/bin/generate_metric_blocklist.py
@@ -47,7 +47,7 @@ while apps:
     for name, info in metrics.items():
         if (
             datetime.datetime.fromisoformat(info["history"][-1]["dates"]["last"])
-            < datetime.datetime(2025, 1, 1)
+            < datetime.datetime(2025, 8, 1)
             and info["in-source"] is False
             and info["type"] != "event"
         ):
--- a/mozilla_schema_generator/configs/glean_v2_allowlist.yaml
+++ b/mozilla_schema_generator/configs/glean_v2_allowlist.yaml
@@ -11,13 +11,15 @@
 # firefox-crashreporter:
 # - metrics
 # - health
-# firefox-desktop:
-# - metrics
-# - health
-# - sync
-# - captcha-detection
-# - new-metric-capture-emulation
-# - pseudo-main
+firefox-desktop:
+- metrics
+- health
+- sync
+- captcha-detection
+- new-metric-capture-emulation
+- pseudo-main
+- usage-reporting
+- newtab
 # firefox-desktop-background-defaultagent:
 # - metrics
 # - health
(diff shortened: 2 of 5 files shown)


bin/generate_metric_blocklist.py
```diff
@@ -47,7 +47,7 @@
     for name, info in metrics.items():
         if (
             datetime.datetime.fromisoformat(info["history"][-1]["dates"]["last"])
-            < datetime.datetime(2025, 1, 1)
+            < datetime.datetime(2025, 8, 1)
```
akkomar, line 50: Shouldn't https://github.com/mozilla/mozilla-schema-generator/blob/main/mozilla_schema_generator/glean_ping.py#L150 be updated too?

akkomar (commented):
Should we also add org-mozilla-fenix-nightly.metrics.1 and org-mozilla-fennec-aurora.metrics.1 to incompatibility-allowlist?

I'm leaving this as a Comment because I'm not 100% sure, happy to approve if I'm wrong.

bin/generate_metric_blocklist.py
BenWu: Yes thanks for catching that

BenWu:
> Should we also add org-mozilla-fenix-nightly.metrics.1 and org-mozilla-fennec-aurora.metrics.1 to incompatibility-allowlist?

I think you're right but unfortunately we would need to delete the tables or redo the table swap, neither of which I want to do. What I'm thinking is just explicitly excluding those two apps. That's my mistake for forgetting to update the metric list before that.

BenWu:
Added a metric blocklist exemption for the two migrated apps which means that the previously removed metrics will be added back. IMO that's fine since they're the test apps and they helped verify that the blocklist works as expected

akkomar (approved):

PR #331, opened 2026-09-22
[Bug 2074273] Remove client_info (and ping_info) from quick-suggest, urlbar-keyword-exposure

scholtzan:
Companion to https://github.com/mozilla/gcp-ingestion/pull/2980

Depends on changes proposed in https://bugzilla.mozilla.org/show_bug.cgi?id=2074273

We'll have to backfill these tables once re-created. We'll likely have to do a small backfill from the error tables as validation will fail for these pings between merging the gcp-ingestion change and updating schemas/recreating the tables


sean-rose (approved):

scholtzan:
It has been decided to include Fenix as in scope, so can we include those fx-suggest / fx-suggest-api pings in both the client patch and the data platform side. Added those to the incompatibility list

